PR #25, opened 2017-06-27
fix(models): add KegDB.toJSON to avoid cycles

dchest:
It was impossible to serialize kegs with JSON.stringify(), because they
contain reference to db, which contains reference to boot keg, which
contains reference to db, resulting in a structure with cycles.

Add toJSON method to KegDB, which will not return boot keg, thus
avoiding the cycle.


anri-asaturov:
There's safe-json-stringify module, but ok ;)

dchest:
Oh, cool, maybe I'll use it if there are any other cycles.

PR #29, opened 2017-08-14
feat(files): unique file paths

seavan:


anri-asaturov:
I don't understand what was changed with this PR

src/models/files/file.js
```diff
@@ -178,8 +178,9 @@ class File extends Keg {
     @computed get cachePath() {
         if (!config.isMobile) return null;
         // we need constant id to find file in cache, but fileId contains some restricted characters
-        const uid = this.name || cryptoUtil.getHexHash(16, cryptoUtil.b64ToBytes(this.fileId));
-        return config.FileStream.getFullPath(`${uid}.${this.ext}`);
+        const uid = cryptoUtil.getHexHash(16, cryptoUtil.b64ToBytes(this.fileId));
+        const name = `${this.name || uid}.${this.ext}`;
+        return config.FileStream.getFullPath(uid, name);
```
anri-asaturov: getFullPath accepts 1 parameter.

seavan:
```    /**
     * Override. Returns full path for file when there's a default cache path implemented in the app.
     * Currently only mobile.
     * @param {string} uid - unique identifier
     * @param {string} name - human-readable file name
     * @returns {string} - actual device path for file
     * @abstract
     * @public
     */
    static getFullPath(uid, name) {
        throw new AbstractCallError();
    }```

now it takes another parameter (uid) which allows for creating unique path (subdirectory with uid is created).

PR #31, opened 2017-08-16
feat: check channel limits on icebear level

seavan:


anri-asaturov:
Also not merging this for same reason as https://github.com/PeerioTechnologies/peerio-desktop/pull/52/files

src/models/chats/chat-invite-store.js
```diff
@@ -184,6 +184,10 @@ class ChatInviteStore {
      * @public
      */
     acceptInvite(kegDbId) {
+        if (chatStore.isChannelLimitReached) {
+            warnings.add('error_acceptChannelInvite');
+            throw new Error('Cannot accept invitation. Channel limit is reached');
```
anri-asaturov: Do not `throw` from function returning promise, caller's 'reject' handlers might never receive this. Instead do `return Promise.reject(new Error('reason for debug'))`

src/models/chats/chat-store.js
```diff
@@ -134,6 +135,20 @@ class ChatStore {
     }
 
     /**
+     * Is number of allowed channels reached for user?
+     * @member {bool} isChannelLimitReached
+     * @memberof ChatStore
+     * @instance
+     * @public
+     */
+    @computed get isChannelLimitReached() {
```
anri-asaturov: Chat store is not the place for this *public* property. All limits are provided by User class.

src/models/user/user.js
```diff
@@ -347,7 +347,7 @@ class User {
         if (this.quota && this.quota.resultingQuotas
```
anri-asaturov: The both channel limit properties ignore current patterns and server quota keg format. Check file quota properties to see what i mean.
- Established pattern is that we return `Number.MAX_SAFE_INTEGER` if quota is null, if first PR would do that, we would have -1 bug instantly :)
- Quota keg is complex and we have to look for specific record in quota arrays for safety.

src/models/chats/chat-store.js
```diff
@@ -436,6 +451,7 @@ class ChatStore {
             this.activate(cached.id);
             return cached;
         }
+        if (isChannel && this.isChannelLimitReached) throw new Error(`Allowed channel limit reached`);
```
anri-asaturov: Again, very optimistic use of 'throw'. In this case, desktop will be stuck in 'waiting' state when limit is reached, that is something that desktop could take care of by catching the exception BUT,
the idea behind icebear api (although maybe not fully implemented) is that clients should not wrap every icebear call in `try{}catch(){}`. If method can throw, icebear should catch it. In this case, right approach is to return null, as client expects chat instance. And then check for null in clients.

PR #32, opened 2017-08-23
feat(ui): don't allow file exceeding current plan limit

seavan:
Because server does not provide "current plan upload limit" feature yet, we use configuration to check for that via activePlans.

That's why I've added server plan ids and all that.

If we decide to get that data from the server, they are to be removed.


src/models/user/user.js
```diff
@@ -391,11 +404,33 @@ class User {
...
+     * @param {number} size - amount of bytes user wants to upload.
+     * @returns {boolean} is file size acceptable for current plan
+     * @memberof User
+     * @instance
+     * @public
+     */
+    canUploadMaxFileSize = (size) => {
+        if (this.isProUser) return true;
+        const adjustedSize = this._adjustedOverheadFileSize(size);
+        if (this.isPremiumUser) return adjustedSize < config.premiumMaxSingleFileUploadSize;
```
anri-asaturov: Why `<` and not `<=`?

src/models/user/user.js
```diff
@@ -391,11 +404,33 @@ class User {
...
+        return adjustedSize < config.basicMaxSingleFileUploadSize;
     };
 
+    @computed get isPremiumUser() {
+        User.current.activePlans.filter(s => config.serverPlansPremium.indexOf(s) != -1).length;
```
anri-asaturov: `config.serverPlansPremium.indexOf(s) != -1` === `config.serverPlansPremium.includes(s)`

PR #33, opened 2017-08-28
feat: take file upload limits from server

seavan:
Slava now provides the following information:

this.quota.resultingQuotas.upload[0].limit

- is null if there are no limits
- is a number if there are limits.


src/helpers/path.js
```diff
@@ -0,0 +1,55 @@
...
+ * @param {string} expression Path expression, e.g. quota.resultingQuotas.upload[metric=maxSize,period=total].limit
+ * @returns {function} Function, which will return undefined or value if evaluation is a success
+ */
+function compile(expression) {
+    if (compileCache[expression]) return compileCache[expression];
+    let result = a => a;
+    expression.split('.').forEach(path => {
+        const prevCaller = result;
+        const found = path.match(/(.*)\[(.*?)\]/);
+        if (found) {
```
anri-asaturov: Add comments inside code describing what and why are we doing

src/helpers/path.js
```diff
@@ -0,0 +1,55 @@
+const compileCache = {};
+
```
anri-asaturov: this file has to be a module of its own (@module) to render correctly in dokumentationjs. Check other similar modules.

src/helpers/path.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Creates a function which evaluates path applied to given object,
+ * step by step. The function will return 'undefined' if any exception
+ * arises during the call
+ * @param {string} expression Path expression, e.g. quota.resultingQuotas.upload[metric=maxSize,period=total].limit
+ * @returns {function} Function, which will return undefined or value if evaluation is a success
```
anri-asaturov: add @public or @protected

src/models/user/user.js
```diff
@@ -416,10 +417,8 @@ class User {
      * @public
      */
     canUploadMaxFileSize = (size) => {
-        if (this.isProUser) return true;
-        const adjustedSize = this._adjustedOverheadFileSize(size);
-        if (this.isPremiumUser) return adjustedSize < config.premiumMaxSingleFileUploadSize;
-        return adjustedSize < config.basicMaxSingleFileUploadSize;
+        const limit = path('quota.resultingQuotas.upload[metric=maxSize,period=total].limit', this);
```
anri-asaturov: change other parts of code to use your ugly path parser to justify merging this :-D

src/helpers/path.js
anri-asaturov: Add tests, so nobody would suffer from bugs in parser or edge cases.

anri-asaturov: rename path.js to something else, 'path' is usually associated with file system.

PR #41, opened 2017-09-11
test(tinydb): fix decryption failure test

dchest:
BTW, why was it changed to just log error and return null instead of throwing?


anri-asaturov:
Because api consumer doesn't care if it throws and why, there's no decision to do based on throw. What's important is if there's a value that's possible to get or not.

PR #42, opened 2017-09-11
feat(db): split TinyDb into manager and collection

dchest:
This allows using many TinyDbs at the same time with different storage
engines.

Global TinyDb is now an instance of manager, which uses
config.StorageEngine as before, but it doesn't deal with key-values
itself, instead it returns collections for TinyDb.user and
TinyDb.system. To open a custom collection, instead of:

new TinyDb(name, encryptionKey)

use

TinyDb.open(name, encryptionKey)

Other methods to open databases (now "collections") were renamed:

TinyDb.openSystemDb is now TinyDb.openSystem,
TinyDb.openUserDb is now TinyDb.openUser.

*Motivation: I want to be able to open Desktop's tinydb from proxy as well as proxy's own tinydb.*


seavan:
@dchest this shouldn't affect our existing client code if we don't use openSystemDb and openUserDb, right?

dchest:
@seavan right, `TinyDb.{system|user}.{set|get}Value` still work the same way.

anri-asaturov:
Btw, guys, i've been marking api access levels as `@public @protected @private`. Do you have any ideas about how to check sdk consumer code (desktop, mobile, proxy) if it tries to access anything but `@public`? Babel plugin? 😱

src/db/tiny-db.js
```diff
@@ -1,147 +1,18 @@
...
-
-    /**
-     * Clears all TinyDb values
-     * @public
-     */
-    clear() {
-        this.engine.clear();
-    }
-}
+const TinyDb = new TinyDbManager(config.StorageEngine);
```
dchest: Oops, getting `TypeError: this.storageEngine is not a constructor` and I suspect this line. `config.StorageEngine` is not yet initialized by the client app, but we already use it (previously it was initialized during first access to `TinyDb.system`). Any ideas on how to fix this, @anri-asaturov?

anri-asaturov:
Case of `S` in `storageEngine`

src/db/tiny-db.js
dchest: One idea is instead of `this.storageEngine` being a constructor, it should be a function returning a constructor. 

```
new TinyDbManager(() => config.StorageEngine)
```

but maybe there's something more elegant.

anri-asaturov: @dchest I'm ok with that, but are you sure it's not a character case issue? 'StorageEngine' vs 'storageEngine'

dchest: No, it takes config.StorageEngine (with big S), and then remembers it in storageEngine property of TinyDbManager.

PR #48, opened 2017-10-13
feat(serverSettings): support for maintenance window

flohdot:
https://peerio.tpondemand.com/entity/6148


src/models/server-settings.js
```diff
@@ -32,6 +32,14 @@ class ServerSettings {
      */
     @observable tag;
 
+    /**
+     * Observable array of timestamps for maintenance begin and end, if applicable.
+     * @member {Array} downtimeMaintenance
```
anri-asaturov: name in `@member` should match the actual name, but don't change it now, we're most likely to remove this kind of declaration in favor of TS soon

flohdot: oh yeah that's just a 🍗-fingers thing

PR #49, opened 2017-10-19
feat(messages): support rich text field in message kegs

lostfictions:
### PR Checklist

- [x] I have merged `dev` into my branch before submitting
- [x] The pull request name follows the standard commit guidelines, e.g. `feat(files): filter by file type`

Simple change to add a rich text field; should be backwards-compatible. PR is a bit noisy since I also turned on/corrected some typings in this branch as I was working (should've been a separate PR, sorry!)

The addition of `@type` fields when there's already a `@member` field is a bit annoying, but Typescript's JSDoc support doesn't look at `@member` (whereas the docs generator does.)


src/models/chats/chat.js
```diff
@@ -676,6 +676,24 @@ class Chat {
...
+     * Create a new Message keg attached to this chat with the given
+     * plaintext (and optional files) and send it to the server.
+     * @function sendMessage
+     * @param {Object} richText - a ProseMirror document tree, as JSON.
+     * @param {string} plaintext - the plaintext representation of the rich text.
+     * @param {Array<string>} [files] an array of file ids.
+     * @returns {Promise}
+     * @memberof Chat
+     */
+    @action sendRichTextMessage(richText, plaintext, files) {
```
lostfictions: Having a separate `sendRichTextMessage` call is a remnant of thinking we'd need the original `text` field AND a new `plaintext` field (and/or would need to do some extra work on the Icebear side) -- maybe `richText` should just be a third optional param to the existing `sendMessage`?

anri-asaturov: Yes, probably a third param, let's revisit this when desktop part is done

PR #54, opened 2017-10-27
Support for writing modules in TypeScript (discussion/WIP, do not merge)

lostfictions:
At Anri's suggestion, today he and I had a call and experimentally configured Icebear to allow writing files in TypeScript; this PR is the result of that call, plus a bit of cleanup afterwards on my part. I think it's worth discussing some of the implications and/or potential issues that may arise from this change.

Further changes to the build system will be needed before this can be merged into dev -- I think Anri's going to handle that.

Some notes for discussion:

- Anri's initial preference was for deprecating Babel in favour of TypeScript as the transpiler, and the current state of this PR/branch reflects this. In the longer term I think this is probably a good idea, but with the way our modules are set up currently it increases the complexity of our build system rather than reducing it. We can set up CI or a prepublish hook so that any consumers of the Icebear NPM package can continue to use it transparently, but anyone linking Icebear locally and working on it will have to remember to spin up the TypeScript compiler in watch mode while making any changes.

  We discussed it a bit more after the call. I think @dchest's suggestion of using the forthcoming [TypeScript support in Babel](https://github.com/babel/babel/pull/5899) maybe makes more sense for the short term since it means nobody's workflow has to change. The caveat there is that we'd need to switch to the Babel 7.0 beta, since it doesn't look like TS support is available via any other channel. (It's also worth noting that this doesn't let us use any extra Babel syntax extensions that aren't available in TypeScript -- though we're not using any right now, which arguably is good for stability and readability.)

  In the longer term, the only functionality in Babel that we wouldn't be able to replicate with a full switch to TypeScript's transpiler is our [`console-kungfu` plugin](https://github.com/PeerioTechnologies/babel-plugin-console-kungfu). The best solution to this is probably to just switch to a logging library with sourcemap support (which would have other benefits too).

- TypeScript and JavaScript files are mixed in the `src/` folder. This seems like the best option -- moving TS files elsewhere would probably just make navigating code a lot more complex and obfuscate where things are coming from. Having them in the same directory means we can also incrementally rename `.js` files to `.ts` when we're ready to switch them over without too much hassle.

- We experimentally switched one file over to TS: `src/helpers/string.js`. It was pretty trivial since it has no dependencies and is all of 5-6 SLOC, but even with that small amount of code we needed to use the `!` (assert non-null) operator in one spot. Generally, "leaf" modules like this (that have no dependencies of their own) seem like a good place to start switching modules over.

- Most instances of `@ts-check` in JS files had to be removed, since we added `"strict: true"` to the `tsconfig.json` file and a bunch of implicit-any and possibly-undefined errors popped up. I don't think there's a way to dial the strictness down for just JS files, so that's worth keeping in mind.

- The `"import/no-unresolved"` eslint rule had to be disabled, since it fails to resolve `.ts` files. (It can theoretically be configured to parse TS files if we set up some config rules and use `typescript-eslint-parser`, but that seemed like a lot of added complexity.)

  In TypeScript proper, imports that fail to resolve are already considered an error -- _if_ you use the `import x from 'module'` syntax. It's possible to continue using the `const x = require('module')` syntax in TS, but it will merrily ignore paths it can't resolve and mark `x` as type `any`, and we really don't want that. (Interfaces and types also can't be imported with the `require` syntax, so it defeats a lot of the benefit of TypeScript.) `import` calls are just transpiled to `require` calls, so there's no real semantic difference here. We should consider banning uses of `require` in TS with a lint rule.

- This PR doesn't currently set up linting for TypeScript files. There are two options: [TSLint](https://github.com/palantir/tslint), or ESLint with [`typescript-eslint-parser`](https://github.com/eslint/typescript-eslint-parser) and [`eslint-plugin-typescript`](https://github.com/nzakas/eslint-plugin-typescript/). I tried the latter option recently and it's still definitely a work-in-progress (notably, rules like `no-undef` throw false positives in some cases, and once in a while the linter spits out a parse error while you're typing) so I think TSLint is the way to go. It doesn't offer the full set of ESLint rules out of the box, but like ESLint it can be extended and folks have contributed packages like [`tslint-eslint-rules`](https://github.com/buzinas/tslint-eslint-rules) to fill the gap.

- Folks working on Icebear will probably want to exclude/ignore the `dist/` directory in their editor in order to avoid accidentally writing code in there and having it get nuked on a recompile. (We currently have the VS Code workspace settings folder `.vscode/` git-ignored so that everyone can keep personal workspace overrides in there, but one option is to un-ignore that and commit it with the correct settings.) This is a non-issue if we keep the Babel pipeline.

- I'm not sure what the state of docs in Icebear is right now -- I haven't looked at them myself, and it seems like most of us just read the code instead of the docs -- but if we want to continue building them with the current system we can point it at `dist/` instead of `src/`. TypeScript preserves JSDoc comments in the output by default, and exported members should keep their names after transpilation. Unfortunately, if we want the docs to have type info for TS files, that info will need to be in both the JSDoc annotations (ie. `@param`, `@returns`, etc.) and in the function signatures.

  This kind of duplication obviously isn't ideal and can easily become the source of errors in docs (though this is already the case, and I've found type errors in our current annotations in Icebear). We could investigate switching to a TypeScript-aware doc generator like [TypeDoc](http://typedoc.org/), but I'm not sure to what extent it supports vanilla JS.

- In general, it seems like having a mixed TypeScript/JavaScript project like this is an added maintenance burden, even if the eventual end result leaves us in a better place. It may be worth estimating how long we'll be in a transitional period and making a more concrete timeline for switching over completely.

cc @dchest @seavan @flohdot


anri-asaturov:
I could probably spend some time learning TS and then convert entire sdk in a week. Took me about 3 days to cover entire codebase with jsdocs.

dchest:
> We could investigate switching to a TypeScript-aware doc generator like TypeDoc, but I'm not sure to what extent it supports vanilla JS.

I use TypeDoc in [a project](https://www.stablelib.com/) (albeit without mixing with JavaScript) and I don't really like it. To be fair, I don't even reference documentation for it or for icebear, since it's so much easier to just navigate through the source code in VSCode and read documentation right in comments there. Does anybody use published icebear documentation or, like me, read it from the source?

> TypeScript and JavaScript files are mixed in the src/ folder. This seems like the best option

👍 

> Folks working on Icebear will probably want to exclude/ignore the dist/ directory in their editor in order to avoid accidentally writing code in there and having it get nuked on a recompile.

Good idea, I sometimes have this issue.


Regarding code style, I'd prefer we use the default style which is applied when you right-click on the text in VSCode and select Format Document, whatever formatting it uses. It would make your type declarations different (without first space), but provides the benefit of not arguing about style and just delegating this stuff to 🤖 formatting robots 🤖 , just like `gofmt` did it for Go code:

```typescript
function getFirstLetter(str: string): string {
    if (!str || !str.length) return '';
    return String.fromCodePoint(str.codePointAt(0)!);
}
```

I think we can even apply the formatting automatically in the commit hook.

anri-asaturov:
Ok let's forget about generating docs for a while, if we focus on providing good quality metadata, doc generation is solvable.

anri-asaturov:
I'm gonna close this keeping the branch for reference. We'll start working on converting everything next week in different branch.

PR #57, opened 2017-11-07
feat: rename oversized to match file property. add oversizeCutoff

seavan:
I have renamed oversized in external images to isOverInlineSizeLimit to match file structure.

I've added:
isOversizeCutoff

to both external image and file. this flag is `true` if we are not do display the image inline under any circumstances. US - https://peerio.tpondemand.com/entity/6873

### DIFF
--- a/src/models/chats/message.js
+++ b/src/models/chats/message.js
@@ -16,7 +16,8 @@ const TaskQueue = require('../../helpers/task-queue');
  * @typedef {{
        url : string
        length : number
-       oversized : boolean
+       isOverInlineSizeLimit : boolean
+       isOversizeCutoff : boolean
    }} ExternalImage
  */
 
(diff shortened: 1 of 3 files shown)


src/models/chats/message.js
```diff
@@ -314,7 +315,8 @@ class Message extends Keg {
         this.externalImages.push({
             url,
             length,
-            oversized: clientApp.uiUserPrefs.limitInlineImageSize && length > config.chat.inlineImageSizeLimit
+            isOverInlineSizeLimit: clientApp.uiUserPrefs.limitInlineImageSize && length > config.chat.inlineImageSizeLimit,
+            isOversizeCutoff: clientApp.uiUserPrefs.limitInlineImageSize && length > config.chat.inlineImageSizeLimitCutoff
```
anri-asaturov: we probably don't want cutoff to depend on user setting, it's mandatory, right?

src/models/files/file.js
```diff
@@ -260,6 +260,10 @@ class File extends Keg {
         return clientApp.uiUserPrefs.limitInlineImageSize && this.size > config.chat.inlineImageSizeLimit;
     }
 
+    @computed get isOversizeCutoff() {
+        return clientApp.uiUserPrefs.limitInlineImageSize && this.size > config.chat.inlineImageSizeLimitCutoff;
```
anri-asaturov: Same here

PR #61, opened 2017-11-09
WIP: cucumber features for files

flohdot:
This is completely WIP scaffolding --I need to read a bunch more about Cucumber to be at all confident about what I'm writing-- but I want some opinions on whether I'm going in the right direction with how I'm structuring it. 

In particular, the volumes section is organized by the role a user has. Does that make sense? 

There's also a lot of interconnectedness between volumes, the file stream, and storage (which already has working tests), and I'm not sure what the best way is to deal with that, especially thinking ahead to actually writing the tests. Any vague hints appreciated.


flohdot:
Maybe more like....

`upload.feature` with:

```
Given I am the owner of the volume
[stuff about uploading]

Given I am an editor of the volume
[stuff about uploading]

Given I am a viewer of the volume
[stuff about how i cannot upload]
```

PR #69, opened 2017-11-13
WIP: Feat file folders

seavan:


dchest:
I don't like that there are `FileFolder` and `FileFolders` classes which differ only by one letter. Can we come up with a better name?

src/models/files/file-store.folders.js
```diff
@@ -6,7 +6,7 @@ const FileFoldersKeg = require('./file-folders-keg');
 const cryptoUtil = require('../../crypto/util');
 const warnings = require('../warnings');
 
-class FileFolders {
+class FileStoreFolders {
```
dchest: 👍

anri-asaturov:
Ok, this branch was prepared for release, no data storage format changes expected, but there are plenty of things to improve, let's review this, folks.

src/models/files/file-folder.js
```diff
@@ -0,0 +1,170 @@
...
+
+    free(file) {
+        if (!this.fileMap[file.fileId]) {
+            console.error('file does not belong to the folder');
+            return;
+        }
+        const i = this.files.indexOf(file);
+        if (i !== -1) {
+            this.files.splice(i, 1);
+            file.folder = null;
```
anri-asaturov: and what about file.folderId ?

src/models/files/file-folder.js
```diff
@@ -0,0 +1,170 @@
...
+
+    addFolder(folder) {
+        if (folder.parent === this) return folder;
+        if (this.folderMap[folder.folderId]) {
+            console.error('folder already exists here');
+            return folder;
+        }
+        if (folder.parent) {
+            console.debug('moving folder from parent');
+            folder.parent.freeFolder(folder);
```
anri-asaturov: So if folder has a parent - we move it, but if file has a parent we do nothing? (line 50)

src/models/files/file-folder.js
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    freeSelf() {
+        if (this.isRoot) return;
+        let root = this;
+        while (!root.isRoot) root = root.parent;
+        this.files.forEach(file => {
+            file.folder = null;
+            root.add(file);
```
anri-asaturov: this calls for `move` method

src/models/files/file-folder.js
```diff
@@ -0,0 +1,170 @@
...
+        });
+        this.files = [];
+        this.folders.forEach(folder => folder.freeSelf());
+        this.folders = [];
+        this.parent && this.parent.freeFolder(this);
```
anri-asaturov: There is a parent because we checked this.isRoot in the beginning. Also i see other usages of this.parent in this file without check for null.

src/models/files/file-folder.js
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    moveInto(file) {
+        if (file.isFolder) {
+            if (this.findFolderByName(file.normalizedName)) {
+                warnings.addSevere('error_folderAlreadyExists');
+                throw new Error('error_folderAlreadyExists');
```
anri-asaturov: Why do we show severe warning AND decide to throw in here? As i see from usage this can break sync forever, for example. We need some consistency in reactions to api methods. Why is our error message in Error constructor a locale key?

src/models/files/file-folder.js
```diff
@@ -0,0 +1,170 @@
...
+            }
+            if (file === this) {
+                console.error('cannot move folder in itself');
+                return;
```
anri-asaturov: See what i mean with previous comment? Here we just silently return. Not even a warning. Need consistency.

anri-asaturov:
I'm merging folder branches because it becomes impossible to wrange all the branches especially having mobile dev pointing to folders branch. Please continue with reviews and fixes.

PR #71, opened 2017-11-21
feat: new 2FA implementation

dchest:
**WIP because I need to test its behavior on the old updated server**

Transparent support for the old server implementation is retained.
It will be removed once the new server is deployed.

https://peerio.tpondemand.com/entity/7210

#### Breaking changes  

The current server will reject authentications due to validation; needs an updated server which will be deployed this week.

Test logins and logouts with and without 2FA, with and without trusted device.

Test need to be performed:

* after we deploy (on staging) updated old server (with old 2fa workflow).
* after we deploy (on staging) new server (with new 2fa workflow).

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


ebilg:
this worked correctly with both staging and treetrunks

dchest:
Thanks @ebilg. Turns out this needs some changes to accommodate for "untrust device" checkbox in logout dialog, so more commits will be coming.

dchest:
Pls re-review.

src/models/user/user.2fa.js
```diff
@@ -83,6 +84,58 @@ module.exports = function mixUser2faModule() {
...
+        return TinyDb.system.removeValue(`${this.username}:twoFACookie`);
+    };
+
+    /**
+     * Returns deviceId from system database.
+     * If there's no deviceId stored, it generates a new one, stores it, and returns.
+     * @returns {Promise<string>}
+     */
+    this._getDeviceId = () => {
+        const key = `${this.username}:deviceId`;
```
anri-asaturov: Why is device id user-scoped?

src/models/user/user.2fa.js
```diff
@@ -83,6 +84,58 @@ module.exports = function mixUser2faModule() {
...
+            if (config.deviceUID && config.deviceUID.length > 0) {
+                h.update(cryptoUtil.strToBytes(config.deviceUID));
+            } else {
+                // We can just use random bytes as deviceId, but
+                // let's follow the same path of getting a hash of it.
+                // It's useful to avoid exposing plain PRNG output to server.
+                h.update(cryptoUtil.getRandomBytes(32));
+            }
+            deviceId = cryptoUtil.bytesToB64(h.digest()); // eslint-disable-line no-param-reassign
+            return TinyDb.system.setValue(key, deviceId).then(() => deviceId);
```
anri-asaturov: `.then(() => deviceId)` can be replaced with `.return(deviceId)`, bluebird ftw

src/models/user/user.2fa.js
dchest: It's different for different users even on the same device: see "Device identifier generation" section here: https://peerio.atlassian.net/wiki/spaces/FD/pages/1868090/2FA+device+management

src/models/user/user.2fa.js
```diff
@@ -83,6 +84,58 @@ module.exports = function mixUser2faModule() {
             });
     };
 
+    /**
+     * Stores 2fa cookie and device trust in system database.
+     * @param {{cookie, trusted}} data
+     * @returns {Promise<void>}
+     */
+    this._set2faCookieData = (data) => {
+        return TinyDb.system.setValue(`${this.username}:twoFACookie`, data);
```
anri-asaturov: System db values are not encrypted, is that ok? We could change workflow to open User db before login and work with it.

src/models/user/user.2fa.js
```diff
@@ -83,6 +84,58 @@ module.exports = function mixUser2faModule() {
...
+            // Not found, generate a new one and save it.
+            const h = cryptoUtil.getHashObject(32, cryptoUtil.strToBytes(this.username), 'PRIDevId');
+            if (config.deviceUID && config.deviceUID.length > 0) {
+                h.update(cryptoUtil.strToBytes(config.deviceUID));
+            } else {
+                // We can just use random bytes as deviceId, but
+                // let's follow the same path of getting a hash of it.
+                // It's useful to avoid exposing plain PRNG output to server.
+                h.update(cryptoUtil.getRandomBytes(32));
+            }
```
anri-asaturov: We we've been abstracting away this kind of operations in crypto.util, let's move this out too.
In crypto.hashing.js, something like `getDeviceId(username, deviceUID)`

src/models/user/user.2fa.js
dchest: It should be ok, since this is device-bound things and 2fa doesn't protect agains device/cookie stealing.

Regarding local db — I wanted to make its key instead depend on some key stored on server, so that you'll need to login to open it to prevent offline bruteforcing of account key with stolen db, but with long account keys this is no longer relevant, so 🤷‍♂️

src/models/user/user.2fa.js
dchest: OK

src/models/user/user.auth.js
```diff
@@ -309,13 +324,44 @@ module.exports = function mixUserAuthModule() {
...
+        // On untrusted devices, during logout we remove
+        // the 2fa cookie, so that the next sign in will
+        // require 2fa.
+        //
+        // On trusted devices, we preserve the cookie,
+        // thus users won't be asked for 2fa code again.
+        return Promise.resolve().then(() => {
+            if (!this.trustedDevice || untrust) {
+                this.trustedDevice = false;
+                return this._delete2faCookieData();
```
anri-asaturov: So we delete 2fa cookie only on explicit signout, right?

dchest: Right. Only on explicit signout and only when the user either didn't mark the device as trusted during 2fa code submission or if it marked it as trusted and then checked the "untrust device" checkbox in logout dialog.

dchest:
Ugh, circular dependency.

anri-asaturov:
@dchest move it to random.js, sorry, not hashing.js

dchest:
Ready for merging.

PR #76, opened 2017-11-23
fix: don't require consumers to have opt installed + ensure all mocha tests run

lostfictions:
the npm script `postinstall` runs not just after running `npm install` in icebear proper, but also when attempting to install it as a dependency. since `opt` is a devDependency, the `postinstall` command will only succeed in this case if the consumer already has `opt` installed -- otherwise the entire install will fail.

eg.
screenshot from 2017-11-22 19 44 38

this PR resolves this issue by softly failing if `opt` isn't present when `postinstall` runs. unrelatedly, it also wraps the `mocha` npm script's globs in quotes so they'll be parsed properly.

...
... 
#### Breaking changes  
...
...
#### Additional notes
... 

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


anri-asaturov:
Are you saying if i get a clean copy i can't nom install?

lostfictions:
quick repro:
```sh
mkdir test-install
cd test-install
npm init
npm install github:PeerioTechnologies/peerio-icebear#v3.3.3
```

anri-asaturov:
wat. dependency postinstall runs after u install it? that's so weird

lostfictions:
incidentally, i recommend looking at
https://github.com/lostfictions/icewrap/blob/master/package.json
and
https://github.com/lostfictions/icewrap/blob/master/build.js
to see everything icebear implicitly forces on consumers

lostfictions:
(and https://github.com/lostfictions/icewrap/blob/66ecdf6dc07d0359196645e5dd0c98430862cecb/index.js#L10-L11, too. and it also sets bluebird as a global i guess?)

PR #79, opened 2017-11-26
feat: retry unfurls failed due to disconnection

anri-asaturov:
This PR tried to detect the reason of unfurl process failure. 
Reminder: unfurl process requests url headers and makes decision based on content-type and length.

Failures due to disconnection previously lead to url being ignored until it appears in another message. Now the unfurl will be re-scheduled to retry after connection is restored.


anri-asaturov:
@ebilg I know this is hard to test, and I did it myself, but you can at least find related TP entities and close them since I made all this changes at once while fixing Alexandra's PR.

src/models/chats/message.js
```diff
@@ -294,10 +295,37 @@ class Message extends Keg {
         }
     }
 
+    _pauseAndRequeue(url) {
+        const queue = Message.unfurlQueue;
+        if (!queue.paused) {
+            when(() => socket.connected, () => queue.resume());
```
lostfictions: I think the logic here is a little tricky... It's "static-like" behaviour even though every message whose unfurling failed when the socket is disconnected will call this method, but the first to get here when the queue isn't paused sets the reaction. Not sure what would improve it at a glance, but it tripped me up.

src/models/chats/message.js
```diff
@@ -294,10 +295,37 @@ class Message extends Keg {
...
+                        this._pauseAndRequeue(url);
+                        return;
+                    }
+                    // There's no reliable way to know if XMLHttpRequest has failed due to disconnection.
+                    // Also, socket.connected is usually updated with a little delay,
+                    // so we rely on this hacky way to postpone the connection check.
+                    // We wait for socket to disconnect and will assume our headers request failed
+                    // due to disconnection if socket disconnects within next few seconds.
+                    // False positives are possible but harmless.
+                    const dispose = when(() => !socket.connected, () => this._pauseAndRequeue(url));
```
lostfictions: This `when` does the same thing as the above, and would fire immediately if the socket isn't connected anyway. I think we can just get rid of the `if (!socket.connected)` branch above and just use this?

(I think `_pauseAndRequeue` should just be inlined if this becomes the only call site as well, for clarity and since it's coupled with the logic happening in here.)

src/helpers/unfurl.js
```diff
@@ -49,9 +49,9 @@ function getContentHeaders(url) {
                 default: break;
             }
         };
+        req.timeout = 30000;
```
lostfictions: Maybe lift this to a constant at the top of the file so it's a bit less of a magic number?

lostfictions:
btw, is there a reason we use a polyfilled XHR instead of isofetch or whatever?

anri-asaturov:
@lostfictions weelllll it's supposed to run in node, chromium and whatever crap hosts of react native under android and ios, so.. wanna experiment with isofetch? :-D

src/models/chats/message.js
anri-asaturov: Yeah i was thinking how to make it better, the unguaranteed event order makes it tricky.
Moving the whole 'pause when not connected' logic to TaskQueue might help, i'll do it some day as other TaskQueue consumers might benefit from it too.

src/helpers/unfurl.js
anri-asaturov: Yeah, the place of this is in app config tbh

src/models/chats/message.js
anri-asaturov: Hmm right, i just didn't want to create reaction when not needed, but you're right, this won't be a bottleneck ever.

seavan:
@anri-asaturov our hosts are not crap! we think different that's it

anri-asaturov:
Manual tests of this are hardly feasible, i've tested it programmatically, merging for tomorrow's staging build

lostfictions:
thanks for the changes!

PR #82, opened 2017-11-27
feat: create and delete image temp cache folder

suprko:

https://peerio.tpondemand.com/entity/6851
https://github.com/PeerioTechnologies/peerio-desktop/pull/157
#### Breaking changes  
n/a
New functions in `node-file-stream` for creating and deleting the temp folder used for downloaded inline images. See desktop PR for specific functionality
#### Additional notes
n/a

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


seavan:
I'll copy US description from TP for @anri-asaturov and anyone else interested:
```
Don’t rely on OS to clean the image cache.

Requirements and edge cases:

Temporary directory should be a unique path, combined of OS temporary directory and a random fs-compatible identifier.

The reason why we need a random identifier is that on Linux, for example, os.tmpdir() is /tmp - shared across other apps. On MacOSX os.tmpdir() always returns a unique identifier.

This identifier is preserved in TinyDB, so that if someday we decide to add a flag “don’t clean up cache on exit”, we could do it easily.

When we launch the app, we check if we have this identifier saved somewhere. If we do, we check if the directory exists.

If it does exist, that means the app wasn’t closed properly or the app couldn’t delete the temporary directory. We try to delete files and remove it from there, and if we fail, it means we don’t have sufficient rights.

If we don’t have rights to clean it up, we generate a new temporary directory identifier, save it to TinyDB and create a directory there.

When we close the app, we catch the mainWindow close event, and unless the directory has been cleaned already, abort the closing process, invoke deleteTempDirectory function and wait for it to complete.

If there’s an error during the execution, we close the app.
```

It's a raw PR, just to clarify if the overall approach is sound.

There are two quirky points here:
- we try to delete last temporary directory when we start the app
- we have to pass temporary directory from main process to web process somehow, so we do it via TinyDB.

dchest:
FYI, OS have special paths for cache directories:

* macOS: `$HOME/Library/Caches/{appName}`
* Linux: `$XDG_CACHE_HOME/{appName}`
* Windows: 🤷‍♂️ 

However, Desktop, being based on Chromium, already stores Chromium caches in Application Support directory, so I think it would be appropriate to store it somewhere there too, in ImageCache.

os.tmpdir() wouldn't be appropriate, since 1) on some Linux it's on a ram disk or may not be encrypted (e.g. when using Ubuntu home folder encryption), 2) people generally have their home folder permissions set properly, but may go wild in /tmp, 3) if we'll have an option to preserve cached images, /tmp it's definitely not a good place.

src/config.js
```diff
@@ -134,6 +134,11 @@ const config = new class {
      */
     ghostFrontendUrl = 'https://';
 
+    /**
+     * Application name
+     */
+    appName = 'peerio';
```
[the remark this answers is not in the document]
seavan: removed. mobile doesn't use this functionality

seavan:
@anri-asaturov check it out again

anri-asaturov:
waiting for @ebilg

src/models/files/node-file-stream.js
```diff
@@ -98,12 +106,67 @@ class NodeFileStream extends FileStreamAbstract {
...
+            try {
+                await this.removeDir(folderName);
+            } catch (e) {
+                console.error(e);
+            }
+        }
+
+        const randomHex = cryptoUtil.getRandomShortIdHex();
+
+        TEMP_ROOT = `${DEFAULT_TEMP_ROOT}/${randomHex}`;
```
lostfictions: use `path.sep` instead of `/` please

dchest: or `path.join(DEFAULT_TEMP_ROOT, randomHex)`

src/models/files/node-file-stream.js
```diff
@@ -3,6 +3,14 @@ const errors = require('../../errors');
 const fs = require('fs');
 const os = require('os');
 const path = require('path');
+const mkdirp = require('mkdirp');
+const rimraf = require('rimraf');
+const config = require('../../config');
+const TinyDb = require('../../db/tiny-db');
+const cryptoUtil = require('../../crypto/util');
+
+const DEFAULT_TEMP_ROOT = path.join(os.tmpdir(), config.appId);
```
dchest: Let's use `path.join(config.StorageEngine.storageFolder, "ImageCache")` (or something else instead of "ImageCache", but not "Cache" because it's taken by Electron) if that's okay for mobile.

src/models/files/node-file-stream.js
```diff
@@ -98,12 +106,67 @@ class NodeFileStream extends FileStreamAbstract {
...
+            mkdirp(folderPath, err => {
+                if (err) reject(err);
+                else resolve();
+            });
+        });
+    }
+
+    static removeDir(folderPath) {
+        return new Promise((resolve, reject) => {
+            rimraf(folderPath, err => {
```
dchest: `rimraf(folderPath, { disableGlob: true }, err => {`

src/models/files/node-file-stream.js
```diff
@@ -98,12 +106,67 @@ class NodeFileStream extends FileStreamAbstract {
...
+                if (err) reject(err);
+                else resolve();
+            });
+        });
+    }
+
+    static async setTempCacheDirectory() {
+        TEMP_ROOT = await TinyDb.system.getValue('tempFolderName') || DEFAULT_TEMP_ROOT;
```
dchest: What's the reason for storing tempFolderName in TinyDb?

If we always derive path from `config.StorageEngine.storageFolder` as suggested earlier, we can get rid of this setter and not store anything in TinyDb.

src/models/files/node-file-stream.js
dchest: Actually, getting `storageFolder` isn't a good idea, instead how about `FileStream` will have a configurable value `cacheFolder` and then desktop and mobile will set it in config just like they do, e.g. desktop:

```
cfg.FileStream = FileStream;
cfg.FileStream.cacheFolder = path.join(app.getPath('userData'), 'ImageCache');
cfg.StorageEngine = StorageEngine;
cfg.StorageEngine.storageFolder = app.getPath('userData');
```

dchest: and if it's not set, just throw an error instead of having a bad default.

seavan:
I refactored everything and simplified things. We do not generate a new unique cache folder each time now, because maybe it's an overkill.

I used the approach @dchest suggested. Not having unique identifier also means not having to store it in TinyDB.

Take a look.

anri-asaturov:
@seavan won't believe it until hear from @ebilg :-D

PR #86, opened 2017-12-01
feat: limit size of temp cache folder

suprko:

https://peerio.tpondemand.com/entity/7067
https://app.clubhouse.io/peerio/story/277/desktop-limit-file-system-cache

desktop https://github.com/PeerioTechnologies/peerio-desktop/pull/167
#### Breaking changes  
n/a
On this branch this app should:
1. (unchanged) Download inline images to temp cache folder
2. When folder is over an arbitrary size limit, delete temp images (oldest first) until folder is below that limit, *except* -
3. Never delete temp images that are in the currently selected chat

#### Additional notes
n/a

----
### Reviewer

* Localisation.
* Architectural decisions.
* Performance.
* Comments.
* Readability.
* DRY.

----
### Repository owner


seavan:
waiting for @anri-asaturov

flohdot:
Hmmm, this is a bit hard to understand (for me) because it's split across files, and I have a feeling someone is gonna be confused by it later on. Can we get some inline comments that explain what's up?

For example, in client-app.js, it would help to have something that tells me why this inlineFileVisible stuff is there.

suprko:
@flohdot added some comments

src/models/client-app.js
```diff
@@ -106,6 +106,20 @@ class ClientApp {
...
+     * @public
+     */
+    inlineFilesVisible = {};
+
+    addFileVisible(fileId) {
+        this.inlineFilesVisible[fileId] = true;
+    }
+
+    removeFileVisible(fileId) {
+        this.inlineFilesVisible[fileId] = false;
```
lostfictions: This smells a bit like a leak -- wouldn't you just want to `delete` the key instead? (Or maybe even better, just use a `Set`?)

src/models/files/file-cache-handler.js
```diff
@@ -0,0 +1,34 @@
...
+     * deleting files (oldest first) that are not currently visible
+     */
+    static cacheDeleteLoop() {
+        for (let i = 0; i < files.length && fileSizeTotal > fileSizeLimit; i++) {
+            const f = files[i];
+            if (clientApp.inlineFilesVisible[files[i].fileId]) continue;
+            config.FileStream.delete(f.tmpCachePath);
+            f.tmpCached = false;
+            fileSizeTotal -= f.size;
+            files.splice(i--, 1);
```
lostfictions: This is a bit too clever to me. There's very, very few valid uses of the prefix increment/decrement operators inline in another statement, and slim-to-none uses of the postfix operators inline. If you're performing two mutations on two variables -- and especially if there's a dependent ordering to those two mutations -- I'd recommend you use two statements for clarity.

lostfictions: incidentally, `splice` is one of the slowest array methods... it's O(n), and it's possibly being called here in a way that's towards the worst-case of n here. plus it's potentially being called for every item in the array, making this O(n^2), every time an image is added to the cache. has this been profiled at all? i could definitely see it causing noticeable extra jank. i also wonder if a proper LRU could make this a lot simpler.

seavan: @lostfictions yup you're right, no place for splice here

seavan:
after discussion with @lostfictions confirmed https://github.com/rsms/js-lru as a no-option - badly maintained leaky js piece of code.

going with node-lru-cache. refactoring.

src/config.js
```diff
@@ -338,6 +338,11 @@ const config = new class {
     };
 
     unfurlTimeout = 30000;
+
+    /**
+     * Maximum total number of cached images which we store before we start deleting the least recent ones
```
lostfictions: number of images?

src/models/files/file-cache-handler.js
```diff
@@ -0,0 +1,33 @@
...
+    length(file) {
+        return file.size;
+    },
+    // remove the item from cache
+    dispose(key, file) {
+        // to avoid solving probable but unlikely complicated situations when total size of visible files
+        // exceeds cache limit, we are deleting file whenever it does become invisible
+        // it may lead to actual cache size exceeding our limit, but this should be a non-issue
+        console.debug(`trying to dispose: ${key}, ${file.tmpCachePath}`);
+        when(() => !file.visibleCounter, () => {
```
anri-asaturov: nice idea

src/models/files/file.download.js
```diff
@@ -117,12 +119,32 @@ function cancelDownload() {
...
+ * @returns {Promise}
+ * @instance
+ * @memberof File
+ * @public
+ */
+function removeCache() {
+    return Promise.resolve((async () => {
+        if (!this.tmpCached) return;
+        try {
+            await config.FileStream.delete(this.tmpCachePath);
```
anri-asaturov: I think delete() will reject if file is already deleted for whatever reason, maybe we should reset tmpCached flag in any case?

seavan:
@ebilg ready for testing

src/models/files/file.download.js
seavan: done

PR #93, opened 2017-12-06
chore: lint copy (find duplicate keys for now)

dchest:
There.


dchest:
I guess we also should turn on linter on CI?

anri-asaturov:
@dchest lint task runs on ci already

PR #112, opened 2018-01-03
test(e2e): file download

dchest:
My first cucumber, so please critique a lot to make it good :)

#### Testing instructions  

No testing.

----
### Repository owner


test/e2e/spec/files/8.download.feature
```diff
@@ -0,0 +1,15 @@
...
+
+    Background:
+        Given I create an account
+
+    Scenario: Download a 1024-byte file
+        Given I upload a 1024 byte file
+        And I see it in my drive
+        Then I can download it
+
+    @long
```
alexandra-marin: This tag doesn't seem to be used anywhere

test/e2e/code/steps/files.js
```diff
@@ -0,0 +1,31 @@
...
+    });
+
+    Then('I can download it', { timeout: 100000 }, async function() { // eslint-disable-line
+        curfile.downloadedName = getTempFileName();
+        await curfile.keg.download(curfile.downloadedName);
+        const downloadedHash = await calculateFileHash(curfile.downloadedName);
+        downloadedHash.should.equal(curfile.hash);
+    });
+
+    After(function() { // eslint-disable-line
```
alexandra-marin: `After` hook can be moved in `hooks.js`.

Right now it will run after every test, but if you add a tag like `@fileCreated` to your scenarios and modify it like so: 
`After('@fileCreated', function() {...` 
it will run only after the scenarios with that tag.

test/e2e/spec/files/8.download.feature
dchest: Yes, I've added it for future in case we'll need a faster way to run tests (since 10 MB upload/download can be slow on some connections.)

test/e2e/code/steps/files.js
dchest: Thanks! I was indeed confused that it was running after every test :)

dchest: Ah, I can't move it to `hooks.js`, because it references local variable. What's the best practice here: move variable into `this` or leave `After` in the file? (I'd prefer the latter, since it preserves the context — it's easier to see cleanup for scenario in the same file.)

alexandra-marin: Yes, best to leave it where it is then.

alexandra-marin: Why it ran after every test: Cucumber is configured to look for step definitions in the whole `test/e2e/code` folder (because some steps are shared or needed in multiple places). If it finds multiple `Before` / `After` hooks in this folder, it runs them all.

dchest: Makes sense, thanks for explanation! I think it's fine then without adding tags since I make sure that After runs only if curfile is defined.

alexandra-marin: Although it works and the null check is helpful, adding tags is more elegant because:
- other scenarios that don't create a curfile won't run an unrelated step
- running more steps than necessary makes it harder to debug and increases the places you have to check for errors when they happen

test/e2e/code/helpers/files.js
```diff
@@ -0,0 +1,96 @@
...
+ * Creates a temporary file filled random data and
+ * returns an object containing file name and hex-encoded
+ * SHA-1 hash of the file.
+ *
+ * @param {number} length file length in bytes
+ * @param {string} [suffix] optional file suffix ('.tmp' by default)
+ * @returns {Promise<{ name: string, hash: string }>}
+ */
+const createRandomTempFile = (length, suffix) => {
+    return new Promise((resolve, reject) => {
```
anri-asaturov: wanna try async/await?

test/e2e/code/steps/files.js
```diff
@@ -0,0 +1,31 @@
...
+    When('I see it in my drive', { timeout: 100000 }, function() {
+        return this.waitForObservable(() =>
+            !curfile.keg.uploading && curfile.keg.readyForDownload,
+        100000);
+    });
+
+    Then('I can download it', { timeout: 100000 }, async function() { // eslint-disable-line
+        curfile.downloadedName = getTempFileName();
+        await curfile.keg.download(curfile.downloadedName);
+        const downloadedHash = await calculateFileHash(curfile.downloadedName);
```
anri-asaturov: Wouldn't it be faster to just compare byte by byte?

test/e2e/code/steps/files.js
anri-asaturov: This should be solved in more universal manner. 
1. Add 'filesToCleanup' array to the Peerio world in here https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/test/e2e/code/world.js
2. Add code that will delete those files in here: https://github.com/PeerioTechnologies/peerio-icebear/blob/dev/test/e2e/code/hooks.js#L14 (before disposing the app instance)

test/e2e/code/steps/files.js
```diff
@@ -0,0 +1,31 @@
...
+
+defineSupportCode(({ When, Then, After }) => {
+    let curfile;
+
+    When('I upload a {int} byte file', async function(int) {
+        curfile = await createRandomTempFile(int);
+        curfile.keg = this.ice.fileStore.upload(curfile.name);
+    });
+
+    When('I see it in my drive', { timeout: 100000 }, function() {
```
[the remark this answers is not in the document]
dchest: Yeah, I copied it from 1.upload.feature :)

Is there a proper .feature file I can refer to for best naming practices? account.feature?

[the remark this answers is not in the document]
anri-asaturov: Yeah I merged file features too soon, my bad. This is a good example: https://github.com/PeerioTechnologies/peerio-icebear/tree/dev/test/e2e/spec/account

test/e2e/code/steps/files.js
dchest: For very big files, probably not, since it would involve reading two files vs reading one file.

anri-asaturov: Maybe smth like this then https://github.com/PeerioTechnologies/peerio-desktop/blob/dev/scripts/store_package_json_hash.js ?

dchest:
Pls re-review.

PR #113, opened 2018-01-03
feat: parallelize chunk download

dchest:
For discussion/initial review.

Needs `config.download.parallelism` in clients set to something more than 1 for proper testing.

<s>(Based off #112, will rebase on top of dev after #112 is merged, so ignore `tests` diff here, discuss it there.)</s>

https://app.clubhouse.io/peerio/story/707/parallelize-chunk-download

#### Testing instructions  

(After review)

* Set various `config.download.parallelism` (1, 2, 4) in the client (in Desktop: add `cfg.download.parallelism = 2`; in `config.js`). We'll probably use 2 initially.
* Download lots of different files, from tiny to very large. Make sure they match exactly with the original — we want to make sure chunks are downloaded correctly and put in the right order (e.g. before uploading test files calculate `shasum uploaded-file` and then after downloading do `shasum downloaded-file` and compare hashes — they must be the same).

----
### Repository owner


dchest:
Rebased on top of dev.

src/config.js
```diff
@@ -187,6 +187,17 @@ const config = new class {
...
+        /**
+         * Parallelism (must be at least 1).
+         *
+         * Note that maxDownloadChunkSize and maxDecryptBufferSize will be
+         * multiplied by parallelism factor.
+         * @member {number} download.parallelism
+         * @memberof config
+         * @public
+         */
+        parallelism: 1,
```
anri-asaturov: it's set to one - so just one download at a time?

src/models/files/file-downloader.js
```diff
@@ -40,78 +41,111 @@ class FileDownloader extends FileProcessor {
...
+        const size = Math.min(this.downloadChunkSize, this.file.sizeWithOverhead - pos);
+        if (size === 0) {
+            this.noMoreChunks = true;
+            this.downloadChain = this.downloadChain.then(() => {
+                this.downloadEof = true;
+                this._tick();
+            });
+            return;
+        }
+        console.log(`Downloading chunk at ${pos} (size: ${size}))`);
```
anri-asaturov: Comment debug line plz, will be too verbose with our inline images.

src/models/files/file-downloader.js
```diff
@@ -40,78 +41,111 @@ class FileDownloader extends FileProcessor {
...
+        this.downloadChain = this.downloadChain
+            .then(() => promise)
             .then(dlChunk => {
+                if (this.stopped) return; // download was cancelled or errored
                 if (dlChunk.byteLength === 0) {
-                    this.downloadEof = true;
-                    this.downloading = false;
-                    this._tick();
-                    return;
+                    throw new Error('Unexpected zero-length chunk');
```
anri-asaturov: Hmm, see that original code didn't throw, I think there was an edge case when we can request a chunk starting from the last byte and server will return 0, it's not an error.

src/config.js
dchest: Yes, it retains the current behavior by default, but client's config can override it.

src/models/files/file-downloader.js
dchest: Previously next chunk size to request was calculated from what chunk size we received, but now we request exactly the chunk size, we can't request 0 length. So it will now throw only if the server returns incorrect size of chunk (not what we requested).

src/config.js
anri-asaturov: Ok, could you add it to desktop plz, so Eren can test.

anri-asaturov:
@ebilg ready to test 
@dchest Prepare desktop branch for Eren please

dchest:
Desktop branch - `feat-parallel-downloads` (https://github.com/PeerioTechnologies/peerio-desktop/pull/204)

PR #127, opened 2018-01-22
feat: add new key on participant leave

anri-asaturov:

https://app.clubhouse.io/peerio/story/2529/rekey-shared-keg-db-on-participant-leave-kick

Note to reviewers: there are not much changes in this PR bcs most of the code for the rekey was written way before, plz follow function calls to review.
 
#### Testing instructions  

@ebilg this needs to be heavily tested (use staging with branch `anri/ch2529/rekey-shared-keg-db-on-participant-leave`)
What changed: every time someone leaves or gets booted from a room - the admin who removes the participant also changes room keys.
So try to break a room by adding/removing participants and invites, changing admins and everything around that. Expected outcome - everything works as before. What can break: fail to decrypt some messages.

@dchest please review and think about how can we avoid unnecessary key changes (when there's no content or no important content with the current key yet)


anri-asaturov:
@ebilg ?

PR #130, opened 2018-01-25
feat: no auto adding tofu for encounters

seavan:

https://app.clubhouse.io/peerio/story/2493/desktop-do-not-add-contacts-without-explicit-action

https://app.clubhouse.io/peerio/story/2492/mobile-do-not-add-contacts-without-explicit-action 

Copying essential info from the task here.

Our contact list is equal to list of our stored Tofu kegs.
In the current implementation we would load Tofu for every contact request.
Also, we would load Tofu twice - as a tofu cache and when loading contact store.

I reworked icebear, introduced separate tofuStore and one entry for tofu loading.

Also, I do not save tofu automatically anymore.

Tofu is saved automatically (and as a result, contact is added) in the following cases:

chat creation
incoming chat deseralization (participants field)
explicitly saving contact (added getContactAndSave function)
In all other cases, Tofu is not saved automatically.


src/models/contacts/contact-store.js
```diff
@@ -364,19 +365,33 @@ class ContactStore {
...
+     * effectively adding it to the contact list
+     * @param {string} username
+     */
+    getContactAndSave(username) {
+        const c = this.getContact(username);
         when(() => !c.loading, () => {
             delete this._requestMap[username];
             if (!c.notFound && !this._contactMap[username]) {
                 this.contacts.unshift(c);
+                // TODO: do we need it here at all?
```
anri-asaturov: Yes, concurrency, if contacts keg loaded before this contact did - the added flag will not get set.

src/models/contacts/tofu-store.js
```diff
@@ -0,0 +1,90 @@
...
+            options: {
+                type: 'tofu',
+                reverse: false
+            },
+            filter: { username }
+        }).then(res => {
+            if (!res.kegs || !res.kegs.length) return null;
+            const keg = new Tofu(getUser().kegDb);
+            keg.loadFromKeg(res.kegs[0]); // TODO: detect and delete excess? shouldn't really happen though
+            // TODO: added caching here
```
anri-asaturov: todo?

seavan: more of a hint. somehow in old implementation the kegs weren't cached after retrieval. was this the intended behaviour, @anri-asaturov ?

anri-asaturov: Yes, it was implemented to always load latest tofu when profile is reloaded. Since updates to tofu wasn't implemented. ATM it's not an issue since we never update tofu. After we add key change feature - it will be an issue.

ebilg:
can be merged

PR #174, opened 2018-04-17
room invite improvements

alexandra-marin:
 https://app.clubhouse.io/peerio/story/3066/sdk-room-invite-improvements
#### Testing instructions  

----
### Repository owner


src/models/chats/chatNotifier.js
```diff
@@ -0,0 +1,23 @@
+const { computed } = require('mobx');
+const chatStore = require('./chat-store');
+const chatInviteStore = require('./chat-invite-store');
+
+class ChatNotifier {
```
anri-asaturov: why is this a separate class and not a part of chat store? there's just 2 members that wrap existing chat store members

alexandra-marin: didn't seemed to belong in any of the stores in particular (chat or invite). also it's a first step to extract UI-related functionality in separate classes and leave stores to contain just business logic.

anri-asaturov: What do you mean, UI-related? Which observable chat store property is not UI related? If something is a first step, we need to know what the other steps will be and where are we going, extracting properties from stores without clear idea behind it doesn't seem right. Draw me a full picture or move this to chat store please.

anri-asaturov: I can show you an example of good module extraction: look at file-store.*.js files in here https://github.com/PeerioTechnologies/peerio-icebear/tree/newfs/src/models/files

alexandra-marin: Some props in chat and invite store are frequently used together in desktop/mobile, which means that they belong together and hence why they were extracted. By UI-related I meant that the way we show the chat list right now is (somewhat) arbitrary and this `chatNotifier` component deals with exactly that. If we ever decide to show chats and invites in another way (and we have in the past), the stores and underlying logic will not need to change. The notifier will get updated and stores will stay clean. Hope I’ve explained it a bit better now

anri-asaturov: They belong together and that's why they have been extracted? Can they belong together in chat store? Or why is that bad?

anri-asaturov: I do understand the concept of breaking down big module to smaller chunks, don't get me wrong. I wanna know the principle. It's called notifier, but it has allRooms property? Does it notify of rooms? Why not move more props in there? Why leave halfway? 
Again, I'm PRO extracting good modules, It needs to make sense.

anri-asaturov: For example, let's find all properties of chat store that sdk never uses, and extract them to some UI props module.
Sorry for being picky, but I think you can agree that when you do a refactor, you have to complete it, or leave things as they wore, otherwise we'll end up with a bunch of half-done refactors.

alexandra-marin: Will move in chat store and open another PR for the refactor

flohdot: (seriously late to this, sorry) re: refactor I think part of the issue is naming. We couldn't think of a good name, but the idea behind chatNotifier was to start the extraction process by removing the things that are displayed in a chat-like way in the UI but aren't actually really/entirely about chats, such as chat invites. There is soon going to be the placeholder DM (ch5591) which also needs a place to go, and in that case it seems even more wrong to put it in the chatstore. The only reason they're in there is because we've made the arbitrary decision to display these things in a chat-like way in the UI, and we could undo that particular decision in the future (and/or adapt our sdk for use by non-UI consumers) and end up with some nonsensical arrangement of code. 

Maybe the `chatNotifier` should be `chatAndChatLikeObjectNotificationHelpers` and `allRooms` should be `allThingsThatAreDisplayedLikeRooms` or `roomsAndRoomIshes` :-P I'm joking, but suggestions are welcome because the naming in an eventual extraction should be clear. 

ANYWAY. for the later refactor, @anri-asaturov, do you have suggestions for what "logical" modules to break down the chat store into?

anri-asaturov: As for naming we can use submodule notation that's already been used, like file-store.folders.js

I don't have complete vision yet, no,  but extracting things that are supposed to be used only on UI has some virtue in it, although, practically it would be better to extract API surface interface. It can be all the things that are visible to external api consumer, but some of them might be used internally too.

anri-asaturov: First priority should be to remove singletons.

ebilg:
can be merged

PR #184, opened 2018-04-25
fix: multiple digest and not only issues

anri-asaturov:
#### Testing instructions  

----
### Repository owner


src/models/update-tracker.js
```diff
@@ -263,24 +284,63 @@ class UpdateTracker {
...
+            this.updated = true;
         } catch (err) {
             if (err && err.name === 'TimeoutError') {
                 this.loadDigest();
             }
         }
     }
 
+    // In the beginning of session, any unread digest items with newKegsCount = 0
+    // or with newKegsCount>0 but for keg types which counters are not useful to us - are leftovers that we can
```
flohdot: what are these items in practice?

anri-asaturov: Digest items, like information about unread read_receipt or any other keg type. 
Basically, when we start a session - we don't rely on digest, we load things from scratch. UNLESS it's a new keg counter that we use (unread message count) - it's safe (and it's better) to mark those digest items as 'read'. Otherwise if your client is never in a situation to mark them as read naturally (for example you never enter some DM where you have unread receipts) - those items are gonna be in your unread digest forever, and it will keep growing.

flohdot: ok yeah, makes sense

src/models/update-tracker.js
```diff
@@ -292,8 +352,7 @@ class UpdateTracker {
                 newKegsCount: 0
             };
             this.digest[id][type] = digest;
-        }
-        if (digest.knownUpdateId >= updateId) return;
+        } else if (digest.knownUpdateId === updateId) return;
```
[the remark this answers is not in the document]
anri-asaturov: mmm.. technically, it can't happen anymore, bcs knownupdateid is updated by the same session that reports updateId, but you're right - it won't hurt leaving it as >=, it might save us from some store implementation bugs

anri-asaturov:
@flohdot the reason was that both digest update and store(chat, any data) update were triggered by socket.authenticated event. So stores would grab old digest(unless somehow it managed to update first), before it manages to update and fail to see the new information coming. And initial digest load and reload after reconnect doesn't not trigger update events, bcs it's, well, not a digest event and it was confusing to do so.

Now all stores are counting on digest load event instead, which is way safer and simplifies logic and debug.

anri-asaturov:
This change actually also might have sped up initial chat list load time, bcs we can skip cycling through digest in the end of loadAllChats - we're sure there's nothing new now, cuz we waited for it to load first.

anri-asaturov:
Another curious fact is that now we basically have 3 steps of catching up after reconnect.
There's socket.preauthenticated which triggers loading of low-level critical stuff like SELF boot keg.
Then socket.authenticated which triggers processed not involved in digest and digest update
and the last is tracker.updated to trigger kegs update.

Will use this with refactor to make steps more streamlined.

PR #187, opened 2018-05-02
fix: markAllAsSeen only if isReadingNewestMessages

suprko:
https://app.clubhouse.io/peerio/story/3578/desktop-adequately-communicate-to-user-about-folder-sharing-in-progress

#### Testing instructions  
see desktop PR

----
### Repository owner


anri-asaturov:
You sure this fixes the issue? You linked the wrong CH story btw

suprko:
I couldn't repro the bug itself originally, so this is a bit of a guess... Need Eren to check if it actually works. I looked at all uses of `markAllAsRead` and couldn't see what else it could be.

anri-asaturov:
Well, this IS the right thing to do for when you have >5 unread messages, cuz there's a delay. But I doubt if fixes the issue in question.

PR #193, opened 2018-05-15
feat: add originalUploadPath to not redownload things when opening

seavan:

https://app.clubhouse.io/peerio/story/5462/mobile-android-upload-image-to-chat-and-try-opening-the-image-from-inline-options-does-nothing

### DIFF
--- a/src/models/files/file.js
+++ b/src/models/files/file.js
@@ -130,6 +130,7 @@ class File extends Keg {
      * @public
      */
     @observable progressMax = 0;
+
     /**
      * currently mobile only: flag means file was downloaded and is available locally
      * @member {boolean} cached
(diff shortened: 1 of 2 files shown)


src/models/files/file.js
```diff
@@ -139,8 +140,39 @@ class File extends Keg {
...
+    /**
+     * File was uploaded in this session from this device
+     * and we saved it's original upload path. Useful for preview
+     * launch
+     * @member {String} originalUploadPath
+     * @memberof File
+     * @instance
+     * @public
+     */
+    @observable originalUploadPath;
```
anri-asaturov: not gonna work with chat files btw, when sharing file in chat, 2 kegs are created in ram

PR #224, opened 2018-06-11
feat: add medcryptor roles to contact profile

alexandra-marin:
 https://app.clubhouse.io/peerio/story/7834/desktop-remove-clerk-s-ability-to-promote-doctor-to-admin
#### Testing instructions  

----
### Repository owner

### DIFF
--- a/test/e2e/code/world.js
+++ b/test/e2e/code/world.js
@@ -95,8 +95,12 @@ class PeerioAppWorld {
         return this.createAccount(username, email, true);
     }
 
-    createMedcryptorAccount = async (medcryptorData) => {
-        return this.createAccount(null, null, false, medcryptorData);
+
+    findContact = async (query) => {
+        const contact = ice.contactStore.getContact(query);
+        await this.waitFor(() => contact.loading === false);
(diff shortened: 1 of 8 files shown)


test/e2e/code/world.js
```diff
@@ -95,8 +95,11 @@ class PeerioAppWorld {
         return this.createAccount(username, email, true);
     }
 
-    createMedcryptorAccount = async (medcryptorData) => {
-        return this.createAccount(null, null, false, medcryptorData);
+    findContact = async (query) => {
```
anri-asaturov: Look, we do this almost every PR, you put things in world.js, I say world.js should be limited to a very small amount of crucial logic that will be used in every scenario, and then you move it back to where it belongs :-D

Note that if you do have a helper that you want to extract and use in a fairly large amount of scenarios that happen to be placed in different .js files - you can do so, create a separate helper file to import from all the places it's needed in. There are already helpers like that implemented (/code/helpers/file.js for example).

ebilg:
did not work

ebilg:
can be merged AFTER TODAYS RELEASE

PR #236, opened 2018-06-22
feat: new contact color calculations

suprko:
CH https://app.clubhouse.io/peerio/story/8090/client-sdk-update-default-avatar-colors
PeerUI https://github.com/PeerioTechnologies/ui-library/pull/4
desktop https://github.com/PeerioTechnologies/peerio-desktop/pull/351
 
#### Testing instructions  
Design has requested new avatar for when user has:
* instead of being totally random, we select from 36 preset colours that are consistent with design's guidelines
* `firstLetter` shows either black or white text depending on the color

This is a breaking change because `color` no longer returns the `#000000` string, instead it returns an object that contains `{value: '#000000', isLight: true}`. So we'll need to update desktop, PeerUI, and mobile all at the same time.

The alternative is to create a new `colorObject`, and then change computed `color` to check `colorObject`, and create new computed `colorIsLight` which also checks `colorObject`. This would be a non-breaking change, but imo would be uglier on icebear.

----
### Repository owner


src/models/contacts/contact.js
```diff
@@ -113,8 +114,9 @@ class Contact {
      * @type {string}
      */
     @computed get color() {
-        if (!this.signingPublicKey) return '#9e9e9e';
-        return `#${cryptoUtil.getHexHash(3, this.signingPublicKey)}`;
+        if (!this.signingPublicKey) return { value: '#e0e1e6', isLight: true };
+        const int = cryptoUtil.getByteHash(1, this.signingPublicKey)[0] % ContactColors.length;
+        return ContactColors[int];
```
anri-asaturov: If our colours are random and very limited now, there's no sense in wasting CPU for hash calculation anymore.
We got to come up with stable and cheap derivation or store avatar colour in profile.
@dchest any ideas? maybe Math.random with username seed?

seavan: @anri-asaturov why? just cache hash somewhere. having Math.random will make it look different in different clients and/or after reinstalls.

i am not even going into that colors may change later on.

anri-asaturov: We don't have caches yet, and when we have it's also not that effective to do excessive computations. We can easily switch to cheaper checksum/hash algo.
With Math.random I thought it's gonna produce same result if you provide same seed, but it seems JS doesn't even support seeds.

dchest: @anri-asaturov I agree, no need for BLAKE2 here. Just use some trivial quick non-crypto hash like this: https://github.com/dchest/cryptopass/blob/master/popup.js#L858-L875

dchest: oh! I thought we get the color from name, but actually, since `this.signingPublicKey` is already a randomish value, we can just get the first byte from it and do `% ContactColors.length`, no need for any hashing at all.

@suprko:

```
-        const int = cryptoUtil.getByteHash(1, this.signingPublicKey)[0] % ContactColors.length;
+        const int = this.signingPublicKey[0] % ContactColors.length;
```

ebilg:
seems to be working, mobile needs to be ready before it can be merged

ebilg:
can be merged

PR #244, opened 2018-07-03
chore: add benchmarks

dchest:
Currently benchmarking some crypto only.

Also disables no-await-in-loop rule for ESLint, because it's a bad rule.

#### Testing instructions  

----
### Repository owner

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,7 +12,8 @@ module.exports = {
         'no-void': 0,
         'prefer-arrow-callback': 0,
         'no-debugger': 0,
-        'prefer-destructuring': 0
+        'prefer-destructuring': 0,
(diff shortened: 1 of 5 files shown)


seavan:
> disables no-await-in-loop rule for ESLint, because it's a bad rule.

really?! so I can use await's in loop?

dchest:
@seavan sure, if you want serial operations. They added this rule because they thought people would assume this performs operations in parallel, like `Promise.all`. But why? 🤷‍♂️

seavan:
*tears of joy*

no more `let promise = Promise.resolve(); for(...) promise = promise.then(...);`

anri-asaturov:
@ebilg this does not affect any existing logic

PR #254, opened 2018-07-15
feat: file, chat and contact cache

anri-asaturov:


src/util.js
```diff
@@ -57,5 +57,26 @@ function tryToGet(fn, defaultValue) {
     return defaultValue;
 }
 
-module.exports = { convertBuffers, formatBytes, tryToGet };
+
+function getCacheDbFullName(name) {
+    if (!name) throw new Error('Cache database has to have a name');
+    const prefix = 'peerio'; // something to separate our databases in case they're in global scope
```
seavan: should we add applabel support here?

src/helpers/performance-polyfill.js
```diff
@@ -0,0 +1,88 @@
+(function(scope) {
```
seavan: it would be nice to have performance features reviewed and merged in a separate branch.

anri-asaturov: What performance features?

src/util.js
anri-asaturov: database is per-app right? so app label is not needed

ebilg:
can be merged

PR #256, opened 2018-07-18
chore: folder filtering functions

suprko:
https://github.com/PeerioTechnologies/peerio-desktop/compare/fix-files-search?expand=1

#### Testing instructions  
Desktop fix (see PR) requires the ability to filter folders by a query passed through argument, rather than through global `fileStore.searchQuery`

----
### Repository owner

### DIFF
--- a/src/models/files/file-store-base.js
+++ b/src/models/files/file-store-base.js
@@ -71,6 +71,12 @@ class FileStoreBase {
             .filter(f => f.normalizedName.includes(q));
     }
 
+    // Filter folders by a query via function argument instead of using this.searchQuery
+    foldersFiltered = (q) => {
+        return this.folderStore.root.allFolders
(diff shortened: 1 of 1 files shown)


src/models/files/file-store-base.js
```diff
@@ -66,7 +66,11 @@ class FileStoreBase {
     // Subset of folders not currently hidden by any applied filters
     @computed get foldersSearchResult() {
         if (!this.searchQuery) return [];
-        const q = this.searchQuery.toUpperCase();
+        return this.foldersFiltered(this.searchQuery);
+    }
+
+    foldersFiltered = (query) => {
```
anri-asaturov: note that generally this is dangerous thing to do, because the result this function returns can only be used from observers or used and disposed quickly, because it's not gonna update with main data source changes.

PR #257, opened 2018-07-18
chore: introduce typescript

lostfictions:

https://app.clubhouse.io/peerio/story/11183/sdk-convert-to-typescript

Upgrades icebear to Babel 7 -- required for `@babel/preset-typescript` -- and introduces the capacity to incrementally rewrite the sdk. One leaf file has been switched over to TS as a test.

TypeScipt `strict` mode is enabled, with `noImplicitAny` disabled.
 
Tests don't work yet -- they use their own `.babelrc` that I haven't switched over. (I think it could probably be changed to just be shared with the root `.babelrc`?)

Use the temporary npm script `test-build` to invoke Babel.

---

To do:
- [ ] Tests (and desktop, and maybe mobile?) use `babel-root-import` (which is just a year-old, deprecated version of [`babel-plugin-root-import`](https://github.com/entwicklerstube/babel-plugin-root-import)) so that needs to be switched over (or de-magicked so we don't need to maintain two different babel files)
- [ ] `eslint-config-peerio` needs to be updated, all local eslint configs probably need to be tweaked. (eslint has been updated locally in this repo, so it currently works -- those changes just need to be brought over to shared config.)
- [ ] every consumer of the library needs to be changed over too
---

#### Testing instructions

I just jumped into this, but we might actually want to do this as a multi-step process:

- First, upgrade to Babel 7 across the codebase, re-run tests, make sure everything works everywhere, maybe compare the output a little by hand if we're feeling paranoid/aren't totally confident about our coverage...
- Then introduce `@babel/plugin-typescript`, change over ESLint, etc. This step is a lot safer than the Babel 7 upgrade in that there's very little risk of breakage or accidental semantic changes. But it involves more work to get everything running, so it's potentially less confusing to do separately.

----
### Repository owner


tsconfig.json
```diff
@@ -3,6 +3,8 @@
     "target": "esnext",
     "experimentalDecorators": true,
     "allowJs": true,
+    "strict": true,
+    "noImplicitAny": false, // FIXME: remove
```
anri-asaturov: it can parse js comments in json?

lostfictions: tsconfig.json supports it. (vs code recognizes this and by default opens it with the "json with comments" syntax.)

lostfictions:
@anri-asaturov Definitely think we should stick with Babel, at least at first. Given our build pipeline, I can't imagine how we'd change desktop, mobile, icebear, etc. over simultaneously...

Unless you're suggesting we don't need to make consumers compile the SDK themselves? That'd be nice since I find that super confusing, but I honestly still don't totally get how/why our build pipeline works (or if I figured it out previously I've forgotten) so I can't comment.

anri-asaturov:
current setup:
- sdk has ./src folder and no ./dist folder
- sdk package.json main == './dist/index.js' 
- desktop runs `"compile:icebear": "babel --source-maps -q app/node_modules/peerio-icebear/src -d app/node_modules/peerio-icebear/dist",` as one of the build steps
- mobile runs `"icebear:compile": "babel node_modules/peerio-icebear/src -d app/lib/peerio-icebear"` (same thing, different destination for some reason)

anri-asaturov:
If we want sdk to precompile sources (at least in 2 different sets of babel plugins, because mobile supports waaay less modern js features), where would you put the compiled sources?

lostfictions:
I think it might make sense to make mobile the exception rather than the rule? So we bundle both a built `dist/` and the raw `src/` in the published icebear package, most consumers can just pull it in without any ceremony, and mobile can have its own build step as just a performance optimization.

seavan:
mobile compiles into app/lib for two reasons:
- because packager ignores live changes on anything in node_modules (and this is unconfigurable).
- because packager doesn't properly work with npm link (it crashes).

anri-asaturov:
So you suggest we push dist to git? @lostfictions

lostfictions:
@anri-asaturov Why not just start publishing an NPM package again?

anri-asaturov:
On every commit in every branch? Our dev process is not compatible with npm packages IMO.

lostfictions:
Yeah, I guess it doesn't work well with our current process. Don't we want reproducibility eventually though? What we're doing now doesn't ultimately seem like the best strategy long-term either...

So -- sure, what if we include dist in the repo?

anri-asaturov:
then it will be
- noisy,
- conflicts in generated code
- longer commits
- if something was wrong with last committer setup and dist is corrupt/incorrectly build - you'll loose time figuring it out

What is the issue with the current setup? Given that we are not a public distributable library with hundreds of consumers (which we can be if there's a need - we'll just make npm piblish from master).

flohdot:
> Given that we are not a public distributable library with hundreds of consumers (which we can be if there's a need - we'll just make npm publish from master every 2 weeks).

We might want to start thinking of ourselves as having many consumers sooner rather than later. I welcome all changes that go in the direction of making us more consumable.

anri-asaturov:
@flohdot it can't be done casually, we need to know who the customers are and what they want, how are they gonna use it, will they use it with Peerio server or their own and other questions. It's a project.

anri-asaturov:
My point is that trying to do things for some hypothetical future while making our daily life worse doesn't sound like a good idea. @flohbot

lostfictions:
> then it will be
>
>   - noisy,
>   - conflicts in generated code
>   - longer commits
>   - if something was wrong with last committer setup and dist is corrupt/incorrectly built - you'll loose time figuring it out
>
> What is the issue with the current setup? Given that we are not a public distributable library with hundreds of consumers (which we can be if there's a need - we'll just make npm publish from master every 2 weeks).

@anri-asaturov okay -- i was trying to articulate a difficulty i was seeing with the current task in general, and with a question you were asking in particular (and hopefully figure out a way together to make things easier, less surprising, and a bit more future-proof). maybe i'm too paranoid that swapping out a major part of our build system will have unforeseen consequences and introduce subtle bugs or changes that are difficult to foresee, especially when it has to happen in lockstep across at least three codebases that are de facto tightly coupled.

for what it's worth and in case it's not clear, i'm in favour of improving qol for us by streamlining our process and making things less complex. (this is something i spend a lot of time doing in my own personal projects as well -- i'm always reflecting on how i'm working and trying to figure out if it could be better.) i apologize if it sounded like i was trying to push a specific way of doing things. mostly i'm just trying to listen to myself when something nags me as confusing. (my experience is usually that if something feels that way to me, it probably feels that way to at least a few other people as well -- they just might be too afraid to question it.)

anyway, maybe all this stuff is a non-issue. i unfortunately really don't feel confident about changing over mobile and desktop in tandem -- and i'm mostly unfamiliar with how mobile is set up -- and with my limited hours i might not be the best point person for this task since i might not be around if weird things crop up later. would it be better if someone more confident about how everything fits together took on this task using this pr as a proof-of-concept?

anri-asaturov:
@lostfictions oh no, I didn't mean to be a grumpy russian, I understand your motives the same way. I'm just trying to figure out what the concrete problems are there with the current setup, what are we trying to fix. Because you know, if it's working - don't fix it, and whatever's best practice today - is gonna be an anti-pattern tomorrow.

So the current build is very simple, sdk provides source files, if consumer can't run them with their js engine - it's consumer's responsibility to transpile. It works well for us so far and supports every aspect of our process, if you see concrete issues with it and how to possibly fix them - please do tell.

To merge this branch it is enough to add the `@babel/typescript`(which, If I got it right, will strip types) to both mobile and desktop, it's a simple step, I'll do it right now, and it will work.

anri-asaturov:
https://github.com/PeerioTechnologies/peerio-desktop/pull/381
here it is for desktop, works nicely. Need to fix eslint tho, some plugin is acting out, will update eslint-config-peerio first.

PR #258, opened 2018-07-19
feat: telemetry support

suprko:
desktop https://github.com/PeerioTechnologies/peerio-desktop/pull/377

#### Testing instructions  
Basic functions for sending events to Mixpanel via HTTP API, and helper functions for things that we want to share between mobile and desktop.

----
### Repository owner


src/telemetry/main.js
```diff
@@ -0,0 +1,70 @@
+// Main telemetry functionality is here, most importantly init() and send()
+const config = require('../config');
+const TinyDb = require('../db/tiny-db');
+const { cryptoUtil } = require('../crypto');
+
+const baseUrl = 'https://api.mixpanel.com/track/?data=';
+const baseObj = {
+    properties: {
+        token: '05ee93d5cdb68e0de0709b6c85200c44', // TODO: this is for "Test setup". remember to replace it. env var?
```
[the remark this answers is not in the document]
anri-asaturov: how do we share env vars in icebear with mobile, @seavan? Config is better place for this.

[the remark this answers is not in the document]
seavan: ok, config

src/telemetry/helpers.js
```diff
@@ -0,0 +1,49 @@
...
+function errorMessage(msg) {
+    const messageNames = {
+        error_usernameBadFormat: 'Bad Characters',
+        error_usernameNotFound: 'Username Not Found',
+        error_usernameNotAvailable: 'Name Already Exists',
+        error_invalidEmail: 'Not Valid Email',
+        error_addressNotAvailable: 'Email Not Available',
+        error_fieldRequired: 'Required',
+        error_loginFailed: 'Sign-in Failed'
+    };
```
anri-asaturov: Not critical in this case, but generally you want to move object constant declaration out of this function, cuz it will be re-created every call (well, until js engine decides to optimize it and cache it, but it's better not to count on that all the time).

src/telemetry/main.js
```diff
@@ -0,0 +1,70 @@
+// Main telemetry functionality is here, most importantly init() and send()
+const config = require('../config');
+const TinyDb = require('../db/tiny-db');
+const { cryptoUtil } = require('../crypto');
+
+const baseUrl = 'https://api.mixpanel.com/track/?data=';
```
anri-asaturov: url also goes to config

src/telemetry/main.js
```diff
@@ -0,0 +1,70 @@
...
+        'Version Number': 1 // refers to our own tracker library versioning
+    }
+};
+
+function getUserId() {
+    return TinyDb.system
+        .getValue('uuid')
+        .then(id => {
+            if (!id) {
+                const newId = cryptoUtil.getRandomGlobalShortIdHex().toString();
```
anri-asaturov: do we need random ids in here? cuz we have random device id already

src/telemetry/main.js
```diff
@@ -0,0 +1,70 @@
...
+const baseObj = {
+    properties: {
+        token: '05ee93d5cdb68e0de0709b6c85200c44', // TODO: this is for "Test setup". remember to replace it. env var?
+        'Version Number': 1 // refers to our own tracker library versioning
+    }
+};
+
+function getUserId() {
+    return TinyDb.system
+        .getValue('uuid')
```
anri-asaturov: uuid is too generic, come up with more specific key name.

src/telemetry/main.js
```diff
@@ -0,0 +1,70 @@
...
+const baseUrl = 'https://api.mixpanel.com/track/?data=';
+const baseObj = {
+    properties: {
+        token: '05ee93d5cdb68e0de0709b6c85200c44', // TODO: this is for "Test setup". remember to replace it. env var?
+        'Version Number': 1 // refers to our own tracker library versioning
+    }
+};
+
+function getUserId() {
+    return TinyDb.system
```
anri-asaturov: system? So we don't do it per user, correct?

src/telemetry/main.js
```diff
@@ -0,0 +1,70 @@
...
+            console.error(err);
+        });
+}
+
+async function init() {
+    const uuid = await getUserId();
+
+    baseObj.properties.distinct_id = uuid;
+    baseObj.properties.Device = config.isMobile ? 'Mobile' : 'Desktop';
+    baseObj.properties['App Version'] = config.appVersion;
```
anri-asaturov: Please change all property calls that are agains js naming convention to use `['propname']` notation

src/telemetry/main.js
suprko: There are tracking events on the sign-up and login pages, before user has signed in (may not have an account yet). We could switch to user UUID after login with some rejigging but marketing/design said per-system is fine.

src/telemetry/main.js
suprko: you mean `properties.Device` => `properties['Device']`? prettier autocorrects this back to `.Device`...

anri-asaturov: 🤦‍♂️

src/telemetry/main.js
suprko: is `config.deviceUID` what you're referring to?

anri-asaturov: yes

anri-asaturov: it might be not acceptable to use it, since server knows who you are by that id

suprko: OK yeah, we need a new random ID I think

src/config.js
```diff
@@ -159,6 +159,22 @@ const config = new class {
...
+         */
+        baseUrl: 'https://api.mixpanel.com/track/?data=',
+
+        /**
+         * Token for our Mixpanel account.
+         * TODO: this is staging token, change to prod.
+         * TODO: or, get from server?
+         * @type {string}
+         */
+        token: '05ee93d5cdb68e0de0709b6c85200c44'
```
suprko: @flohdot here you go 🤷‍♂️

src/telemetry/main.js
```diff
@@ -0,0 +1,65 @@
...
+const TinyDb = require('../db/tiny-db');
+const { cryptoUtil } = require('../crypto');
+
+const baseObj = {
+    properties: {
+        'Version Number': 1 // refers to our own tracker library versioning
+    }
+};
+
+async function getUserId() {
```
suprko: @seavan is this what you meant?

src/telemetry/strings.js
```diff
@@ -0,0 +1,62 @@
+const S = {
```
suprko: OK so.... in my opinion we keep this as-is for now. Firstly, we want to release next week, and grep-ing all of the strings right now would make me sad. Secondly, if (when?) we change to sending different events to different tokens, the entire event property structure is likely to change. Like we probably won't have "location" props for event flows that all take place in the same location, for example. So I think we can decide to reorganize these strings when we have a better sense of how the events themselves will be restructured.

src/telemetry/main.js
dchest: Yes, we'd like tracking ids to be different from deviceUIDs, so it needs to be new.

src/telemetry/main.js
```diff
@@ -0,0 +1,65 @@
...
+        await TinyDb.system.setValue('telemetryUserId', newId);
+        return newId;
+    }
+    return userId;
+}
+
+async function init() {
+    const uuid = await getUserId();
+
+    baseObj.properties.distinct_id = uuid;
```
dchest: can just 

`baseObj.properties.distinct_id = await getUserId();`

src/telemetry/main.js
```diff
@@ -0,0 +1,65 @@
...
+    properties: {
+        'Version Number': 1 // refers to our own tracker library versioning
+    }
+};
+
+async function getUserId() {
+    const userId = await TinyDb.system.getValue('telemetryUserId');
+
+    if (!userId) {
+        const newId = cryptoUtil.getRandomGlobalShortIdHex().toString();
```
dchest: No need for `toString()` here, it already returns a string.

src/telemetry/main.js
```diff
@@ -0,0 +1,65 @@
...
+    const object = Object.assign({}, eventObj, baseObj);
+    object.properties = properties;
+
+    const data = global.btoa(JSON.stringify(object));
+    const url = `${config.telemetry.baseUrl}${data}`;
+
+    // TODO: this makes it easier to see tracking events for testing purposes. Remove before release.
+    console.log(object);
+
+    window.fetch(url, {
```
anri-asaturov: this is SDK, It doesn't know `window`, use `global-context.js` module

src/telemetry/main.js
```diff
@@ -0,0 +1,65 @@
...
+    baseObj.properties.token = config.telemetry.token;
+    baseObj.properties.Device = config.isMobile ? 'Mobile' : 'Desktop';
+    baseObj.properties['App Version'] = config.appVersion;
+}
+
+function camelToTitleCase(text) {
+    return (text[0].toUpperCase() + text.slice(1)).split(/(?=[A-Z])/).join(' ');
+}
+
+function send(eventObj) {
```
anri-asaturov: wrap the body with `try{}catch(){}` so telemetry fail never affects users

src/telemetry/main.js
```diff
@@ -0,0 +1,65 @@
...
+
+    if (!userId) {
+        const newId = cryptoUtil.getRandomGlobalShortIdHex().toString();
+        await TinyDb.system.setValue('telemetryUserId', newId);
+        return newId;
+    }
+    return userId;
+}
+
+async function init() {
```
anri-asaturov: current code doesn't have a chance to fail, but for future changes in `init` and callees I would wrap this in try-catch too

src/telemetry/main.js
```diff
@@ -0,0 +1,65 @@
...
+    // This song-and-dance merges the properties first, assigns the object, then assigns the object's properties
+    properties = Object.assign(properties, baseObj.properties);
+    const object = Object.assign({}, eventObj, baseObj);
+    object.properties = properties;
+
+    const data = global.btoa(JSON.stringify(object));
+    const url = `${config.telemetry.baseUrl}${data}`;
+
+    // TODO: this makes it easier to see tracking events for testing purposes. Remove before release.
+    console.log(object);
```
anri-asaturov: console.debug is good idea in such cases, one can quicky search for debug strings

src/telemetry/helpers.js
```diff
@@ -0,0 +1,46 @@
...
+    if (location) ret[1].location = location;
+    if (sublocation) ret[1].sublocation = sublocation;
+
+    return ret;
+}
+
+module.exports = {
+    textInput,
+    duration
+};
```
anri-asaturov: Convert to TS please

src/telemetry/main.js
```diff
@@ -0,0 +1,65 @@
...
+
+    window.fetch(url, {
+        method: 'POST'
+    });
+}
+
+module.exports = {
+    init,
+    send
+};
```
anri-asaturov: Convert to TS please

src/telemetry/strings.js
```diff
@@ -0,0 +1,62 @@
...
+    TEXT_INPUT: 'Text Input',
+    TOGGLE_VISIBILITY: 'Toggle Visibility',
+    USERNAME: 'Username',
+    VIA_SYNC: 'Via Sync',
+    VIEW_LINK: 'View Link',
+    WELCOME_SCREEN: 'Welcome Screen',
+    WHERE_ACCOUNT_KEY: 'Where to find Account Key'
+};
+
+module.exports = S;
```
anri-asaturov: Convert to TS please

src/telemetry/main.ts
```diff
@@ -0,0 +1,73 @@
...
+        await TinyDb.system.setValue('telemetryUserId', newId);
+        return newId;
+    }
+    return userId;
+}
+
+export async function init() {
+    try {
+        baseObj.properties.distinct_id = await getUserId(); // eslint-disable-line
+        baseObj.properties.token = config.telemetry.token;
```
seavan: should be serverSettings.mixPanelClientToken

instead of setting the token once it should be retrieved each time in the send function, so we're up-to-date with the server.

src/config.js
seavan: @suprko should be removed in favor of serverSettings.mixPanelClientToken (see my comment below).

mixPanelClientToken is not available immediately, only after socket is connected and initial request is done to retrieve server settings.

it is also re-read each time socket is reconnected, so it makes sense to actually ALWAYS read it from serverSettings in case token is updated on the server.

src/telemetry/main.ts
```diff
@@ -0,0 +1,73 @@
...
+        const object = Object.assign({}, eventObj, baseObj);
+        object.properties = properties;
+
+        const data = g.btoa(JSON.stringify(object));
+        const url = `${config.telemetry.baseUrl}${data}`;
+
+        console.log(object);
+        g.fetch(url, {
+            method: 'POST'
+        });
```
anri-asaturov: Wait, why is it a POST, but data is in url?

suprko: is POST not the right thing to use? for sending encoded data?

anri-asaturov: You're not sending any POST data. What does the api say?

src/telemetry/helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+    error_invalidName: 'Not allowed',
+    error_wrongAK: 'Invalid Account Key'
+};
+
+function errorMessage(msg: string): string {
+    return ERROR_MESSAGES[msg] || msg || 'Unknown Error Type';
+}
+
+// Factory for text input events
+interface textInputProps {
```
lostfictions: Capitalize interface names plz; I'll look into turning a lint on for this

src/telemetry/helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+    state: string;
+    errorType?: string;
+}
+export function textInput(
+    item: string,
+    location?: string,
+    sublocation?: string,
+    state?: string,
+    errorMsg?: string
+) {
```
lostfictions: Not too sure why this function exists. It seems to mostly take a list of parameters and just pack them into an object, and then return a tuple. Couldn't it just take an object and not have to munge any of the parameters itself? (Generally when you have a function with five parameters and four of them are optional, what you probably want is an object anyway.)

This would also help avoid the duplication/lack of DRY in this function signature and the interface right above it.

src/telemetry/helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+    error_addressNotAvailable: 'Email Not Available',
+    error_fieldRequired: 'Required',
+    error_loginFailed: 'Sign-in Failed',
+    error_invalidName: 'Not allowed',
+    error_wrongAK: 'Invalid Account Key'
+};
+
+function errorMessage(msg: string): string {
+    return ERROR_MESSAGES[msg] || msg || 'Unknown Error Type';
+}
```
lostfictions: Are "unknown error types" -- which would entail passing an empty string into this function? -- actually something we want to allow, or is this just a failsafe to try to avoid `undefined`?

It kinda looks to me like this function could be removed and anywhere it's used we just require a `keyof ERROR_MESSAGES` and use it to index into `ERROR_MESSAGES` -- that way error types are actually well-defined, and if any of them are missing we know we need to add them in here.

src/telemetry/helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    if (location) properties.location = location;
+    if (sublocation) properties.sublocation = sublocation;
+    if (errorMsg) properties.errorType = errorMessage(errorMsg);
+
+    return [S.TEXT_INPUT, properties];
+}
+
+// Factory for duration events
+interface durationProps {
```
lostfictions: capitalize

src/telemetry/helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+    item?: string;
+    location?: string;
+    sublocation?: string;
+    totalTime: number;
+}
+export function duration(item: string, location: string, sublocation: string, startTime: number) {
+    const totalTime = Math.round((Date.now() - startTime) / 1000);
+
+    const properties: durationProps = { totalTime };
```
lostfictions: same comment about just taking an object of the type we want. most of this code is just boilerplate and disappears

src/telemetry/main.ts
```diff
@@ -0,0 +1,76 @@
+// Main telemetry functionality is here, most importantly init() and send()
+const config = require('../config');
+const TinyDb = require('../db/tiny-db');
+const { cryptoUtil } = require('../crypto');
+const g = require('../helpers/global-context');
+const serverSettings = require('../models/server-settings');
+
+interface propertyTypes {
+    'Version Number'?: number;
+    distinct_id?: string; // eslint-disable-line
```
lostfictions: we should never be using `// eslint-disable-line` bare -- instead, we should always disable the specific error (or errors) in question, so that we don't accidentally suppress any other error that might emerge at this line later.

in this case i assume it's camelcase? so you can write `// eslint-disable-line camel-case` or whatever the rule name is, and it'll only disable that specific rule instead of turning off _all_ linting for that line. if you need to disable multiple errors you can comma-separate them, ie. `// eslint-disable-line annoying-rule-one, bad-rule-two`. if eslint doesn't recognize an error you're trying to disable it does nothing, so you can (and should) use this to annotate _why_ you're disabling a rule, ie.:
`// eslint-disable-line camel-case, (mixpanel requires this specific property name)`

src/telemetry/main.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const baseObj: { properties: propertyTypes } = {
+    properties: {
+        'Version Number': 1 // refers to our own tracker library versioning
+    }
+};
+
+async function getUserId(): Promise<string> {
+    const userId: Promise<string> = await TinyDb.system.getValue('telemetryUserId');
```
lostfictions: i haven't looked at tinydb too closely, but `getValue` returns a promise that returns a promise?

src/telemetry/main.ts
```diff
@@ -0,0 +1,76 @@
...
+        const newId: any = cryptoUtil.getRandomGlobalShortIdHex();
+        await TinyDb.system.setValue('telemetryUserId', newId);
+        return newId;
+    }
+    return userId;
+}
+
+export async function init() {
+    try {
+        baseObj.properties.distinct_id = await getUserId(); // eslint-disable-line
```
lostfictions: ditto re: not disabling the linter entirely.

src/telemetry/main.ts
```diff
@@ -0,0 +1,76 @@
+// Main telemetry functionality is here, most importantly init() and send()
+const config = require('../config');
+const TinyDb = require('../db/tiny-db');
+const { cryptoUtil } = require('../crypto');
+const g = require('../helpers/global-context');
+const serverSettings = require('../models/server-settings');
+
+interface propertyTypes {
```
lostfictions: uppercase

src/telemetry/main.ts
```diff
@@ -0,0 +1,76 @@
...
+    'Version Number'?: number;
+    distinct_id?: string; // eslint-disable-line
+    token?: string;
+    Device?: 'Mobile' | 'Desktop';
```
lostfictions: looking at this file these seem to actually be required properties, not optional ones? were you just having trouble getting it to compile when building an object piecemeal below?

src/telemetry/main.ts
```diff
@@ -0,0 +1,76 @@
...
+    } catch (e) {
+        console.error('Could not initialize telemetry.', e);
+    }
+}
+
+function camelToTitleCase(text: string): string {
+    return (text[0].toUpperCase() + text.slice(1)).split(/(?=[A-Z])/).join(' ');
+}
+
+export function send(eventObj) {
```
lostfictions: please annotate function types in new typescript files

src/telemetry/main.ts
```diff
@@ -0,0 +1,76 @@
...
+        // So we still write props in camelCase when sending events from client, and convert them here.
+        let properties = {};
+        Object.keys(eventObj.properties).forEach(itemInCamel => {
+            const item = camelToTitleCase(itemInCamel);
+            properties[item] = eventObj.properties[itemInCamel];
+        });
+
+        // `properties` will be overwritten if you directly assign eventObj to baseObj or vice versa.
+        // This song-and-dance merges the properties first, assigns the object, then assigns the object's properties
+        properties = Object.assign(properties, baseObj.properties);
```
lostfictions: generally i'd suggest to use a spread instead of object.assign these days -- i think it's more concise and the intent is clearer, since object.assign sometimes gets used for different things.

src/telemetry/main.ts
```diff
@@ -0,0 +1,76 @@
...
+        const object = Object.assign({}, eventObj, baseObj);
```
lostfictions: this is definitely better expressed as a spread

src/telemetry/main.ts
```diff
@@ -0,0 +1,76 @@
...
+        object.properties = properties;
+
+        const data = g.btoa(JSON.stringify(object));
+        const url = `${config.telemetry.baseUrl}${data}`;
+
+        console.log(object);
```
lostfictions: stray log?

suprko: Anri said to keep it

src/telemetry/main.ts
suprko: Yeah basically. So at the top should it be declared with all these properties empty?

src/telemetry/main.ts
suprko: your exact comment was "console.debug is good idea in such cases, one can quicky search for debug strings", actually did you mean change console.log to console.debug? I thought they were basically the same

src/telemetry/main.ts
lostfictions: In this case the base properties could probably be a `let` and assigned all at once from the `init` function, which to me makes things clearer. or else if you want to make sure the constants like the version number are near the top of the file, then the "constant" base properties could be in their own object declared as a `Readonly<Partial<PropertyTypes>>` (this is kind of a confusing name btw -- maybe just `Properties`?

src/telemetry/main.ts
anri-asaturov: yes, i remember now, i meant that when you output debug logs, use console.debug() so it's easier to find them later.

anri-asaturov: They are the same, but if we would see console.debug in review we would knew for sure it's something forgotten.

suprko: OK so when we actually want to output a log, use console.debug. In this case we can remove it though, right? Don't need to log every single event right?

anri-asaturov: :-D This went way too far.
If you need to log something in production - use console.log
if you're debugging stuff and don't want your log statement to end up in production - use console.debug, that's all. 

What exactly to debug in production is a separate question that I have not answered myself yet :-D

src/telemetry/helpers.ts
lostfictions: Well, if something is marked as optional and isn't actually optional I'd say that's even more of a problem. I just grepped through the desktop telemetry branch and it seems like this is only used in four places so I'm not sure it'd be a huge deal? This pattern seems more error-prone to me, not less, especially if you can lean on interfaces/types to enforce what needs to be in a given object for an event.

suprko: It's used 7 places on mobile, but if we're 100% sure about removing this then I can update those events too

lostfictions: I guess it's your call -- these functions have a real code smell to me, which was why I asked about them even though I didn't have context for how/where they're used. Like I said, I think functions like this with long signatures
`function takesABunchOfRandomProps(thing: string, otherThing?: number, thirdThing?: Dog, yetAnotherThing?: whatever)`
are generally better represented as
`function takesAnObject(config: ThingConfig)` -- that's not a TypeScript thing, that's just a general JS convention (and programming convention; python has `kwargs`, even older languages are trying to move away from functions with long positional lists, etc.)

I think it's a relatively simple refactor and should only take a few minutes even to change at a dozen sites, but I don't have a lot of context for what you're doing so it's up to you.

lostfictions: I wasn't saying you necessarily have to get rid of the functions entirely either -- if you think it's still useful to have a helper that takes a props object and wraps it in the tuple that can be consumed by analytics, maybe that totally makes sense.

suprko: Ehhh if we were passing in an object with manually defined props anyway, all the function would do is add the `S.TEXT_INPUT` event name, at which point the function seems even more superfluous.

I think the way to go is as you said above, having a propery defined `interface` for textinput events, which will take care of the concern that this function was meant to address.

suprko:
Updated. I think I got to everyone's suggestions...?

suprko:
Ready for re-review, I think
@dchest @lostfictions @seavan @anri-asaturov

src/telemetry/main.ts
```diff
@@ -0,0 +1,83 @@
...
+    Device: '' | 'Mobile' | 'Desktop';
+    'Version Number': number;
+    'App Version': string;
+}
+let baseProperties: Properties = {
+    distinct_id: '',
+    token: '',
+    Device: '',
+    'Version Number': 1,
+    'App Version': ''
```
lostfictions: Sorry, this really isn't what I meant :( I probably didn't articulate it well, but you shouldn't be initializing a field with unusable/junk data just to satisfy the typechecker. Doing this throws away a lot of the benefits of typechecking, and particularly strict null checks, but worse is that it's often actually more bug-prone than accidentally forgetting to initialize the field and getting a null ref later.

I know you've been working on this PR for a while and I don't want to further belabour this too much, but I also think this is super-critical to understand -- not just as a typescript thing, but as a core idea of software engineering and incidental complexity/how bugs happen. Happy to discuss the problem and approaches tomorrow morning.

src/telemetry/index.ts
```diff
@@ -0,0 +1,11 @@
+import S from './strings';
+import { init, send } from './main';
+import { duration, errorMessage } from './helpers';
+
+export default {
```
lostfictions: You should almost never be using `export default { x, y, z }` -- just `export` them without `default` instead. The former form is creating a new object, assigning a bunch of named properties to it, and exporting it as default. It prevents destructuring imports, can be confusing to use (since it's inconsistent with how we normally export stuff), and in web contexts prevents tree-shaking.

You can also re-export names directly from another file, instead of importing them and then re-exporting them: https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export

@anri-asaturov Might be worth enabling this lint? https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-anonymous-default-export.md

src/telemetry/main.ts
```diff
@@ -0,0 +1,71 @@
+const config = require('../config');
+const TinyDb = require('../db/tiny-db');
+const { cryptoUtil } = require('../crypto');
+const g = require('../helpers/global-context');
+const serverSettings = require('../models/server-settings');
```
anri-asaturov: require -> import

ebilg:
can be merged

PR #259, opened 2018-07-23
fix(chat): don't report initial page load until we added messages

dchest:
This makes initial chat load indicator spin until we added messages and
avoids many renders of intermediate state while messages are being added
on desktop.

#### Testing instructions  

Load chats, see if they load 🤷‍♂️ 

----
### Repository owner


dchest:
`Promise.finally` is bluebird, do we have it on mobile?

anri-asaturov:
yes, we have it everywhere, even async functions return bluebird promises due to babel plugin wrapping

PR #260, opened 2018-07-23
fix: lowercase all invitation emails [ch11438]

seavan:
https://app.clubhouse.io/peerio/story/11438/mobile-invited-users-should-not-be-case-sensitive


anri-asaturov:
This should be dealt with server side.

PR #262, opened 2018-07-24
refactor: make Keg.loadFromKeg async

dchest:
Minimal changes to make loadFromKeg async.

#### Testing instructions  

This touches every keg load, so *everything* needs to be tested: login, rooms, DMs, files, etc., and the order of messages.

----
### Repository owner


src/models/chats/chat.message-handler.js
```diff
@@ -126,7 +126,7 @@ class ChatMessageHandler {
                 this.setDownloadedUpdateId(resp.kegs);
                 this.markAllAsSeen();
                 console.log(`Got ${resp.kegs.length} updates for chat`, this.chat.id);
-                this.chat.addMessages(resp.kegs);
+                await this.chat.addMessages(resp.kegs);
```
dchest: @anri looks like `await` here isn't needed? `addMessages` was returning Promise even before.

anri-asaturov: yeah it does return a promise

dchest:
Question about `addMessages` queue. I kept it as-is:
```
     @action addMessages(kegs, prepend = false) {
         if (!kegs || !kegs.length) return Promise.resolve();
         return new Promise((resolve) => {
             // we need this because we don't want to add messages one by one causing too many renders
             const accumulator = [];
             for (let i = 0; i < kegs.length; i++) {
                 this._addMessageQueue.addTask(this._parseMessageKeg, this, [kegs[i], accumulator]);
             }
             this._addMessageQueue.addTask(this._finishAddMessages, this, [accumulator, prepend, kegs], resolve);
         });
     }
```

but `this._parseMessage` is now async. Is it okay for the queue? I mean, I checked that it works, but not sure if it works as intended.

anri-asaturov:
Yes, it needs to be a queue, bcs queue is throttled and mobile actually has some small pause configured to avoid blocking UI when loading pages of messages.

dchest:
Ugh:  https://mobx.js.org/best/actions.html#async-await

Will need to insert lots of `runInAction` there 😨

anri-asaturov:
@dchest messages render in batches (pushed to array synchronously) so it's fine

src/models/mail/ghost-store.js
```diff
@@ -53,15 +53,15 @@ class GhostStore {
         if (this.loading || this.loaded) return;
         this.loading = true;
         this._getGhosts()
-            .then(action(resp => {
+            .then(action(async resp => {
```
dchest: @anri-asaturov should I remove `action` from here (and everywhere else there's async) then? Because it will only work until the the first `await`.

dchest: OK, I'll then leave it as is, it would be easier to refactor if we have them as "hints".

dchest:
Colleagues, let's not reformat all files in the project while there are open PRs :) Because I didn't want to re-read all the code I changed to manually resolve conflicts, I had to undelete PR #257 branch, fetch it, rebase it to delete `chore: prettier` commit, merge it with this branch, then do the prettier re-formatting, then merge `dev`, and resolve conflicts by selecting incoming changes 
fun

anri-asaturov:
😱

lostfictions:
@dchest oof, really sorry about that :(  i thought that branch wouldn't get merged for another week, and that we would just merge dev into it in the meantime...

it's obviously also kind of not ideal to merge in a branch that adds a new test that can fail the ci and swaps out a chunk of our tooling without first checking in with everyone and making sure they're on board :(

anri-asaturov:
Didn't know you both are serious.
When is a good time to do a full reformat? Never, there are always branches under development, there are always PRs. For what is worth Friday was the best time cuz we just merged all we could and made a release, in 1 week there will be more branches and more code in them then today.

dchest:
What I meant is that since we combine commits into one when merging to dev, it would have been a good idea to make the prettier commit separate so that that other changes could be easily merged without conflict separately. That's basically what I did by first merging the deleted branch without this commit, and then doing a separate prettier commit over those changes.

ebilg:
can be merged

PR #265, opened 2018-08-01
feat: rekey shared keg DB on participant leave [ch2529]

dchest:

https://app.clubhouse.io/peerio/story/2529/rekey-shared-keg-db-on-participant-leave-kick

Now waits for key in boot keg to appear for at least 30 seconds. Time out is here to prevent someone from manually sending kegs that have non-existing keyIds. (Not sure if 30 seconds is too long or too short).

For testing, I disabled a trivial quick path, will re-enable after merging.

There's some code duplication, not sure if we should abstract it away.
 
#### Testing instructions  

It's hard to test the issue this code is solving, since it depends on server and connection. General testing: kegs should decrypt, chats should have new keys after a participant leaves. To test the edge cases, we'll probably need to repeat a trivia with kicking out participants that revealed the issue.

----
### Repository owner


anri-asaturov:
30 sec is a bit short, yes, even request timeout is 60, 
let's go with 2 minutes. 
Also, want to try to write e2e test for this?

src/models/kegs/boot-keg.js
```diff
@@ -18,6 +19,31 @@ class BootKeg extends Keg {
...
+     * @param {string} keyId
+     * @param {number} timeout
+     */
+    async getKey(keyId, timeout = 30000) {
+        /* TODO(dchest): uncomment after testing
+        if (this.keys[keyId]) {
+            // quick path
+            return this.keys[keyId];
+        }
+         */
```
anri-asaturov: should we invoke async get only if there's no key present? quck path will be the case 99% of time and we add one more async call to it.

dchest:
Increased timeout.

I was thinking about tests, yes. It kinda lies in the boundary between e2e test and unit tests, since to check test results requires digging into internals. But it should have a test, so I'll try to write it.

src/models/kegs/boot-keg.js
dchest: Agree, I'll better move the quick path into loadFromKeg without async.

src/models/kegs/boot-keg.js
```diff
@@ -18,6 +19,31 @@ class BootKeg extends Keg {
...
+        let resolve;
+        const promise = new Promise(_resolve => {
+            resolve = _resolve;
+        });
+        const disposeReaction = when(() => this.keys[keyId], resolve);
+        await promise.timeout(timeout).catch(() => {
+            disposeReaction();
+        });
```
anri-asaturov: check helpers/prombservable.js (asPromise)
this can be
```
await asPromiseNegative(this.keys, keyId, undefined).timeout(..)...
```

dchest: I looked into it, but timeout won't dispose of `when` if I use it.

ebilg:
can be merged

PR #266, opened 2018-08-01
feat: switch to first chat if removed from active chat [ch11920]

seavan:
https://app.clubhouse.io/peerio/story/11920/desktop-do-not-switch-chat-when-removed-from-room-if-already-in-a-different-chat


ebilg:
can be merged

PR #267, opened 2018-08-01
fix: when receiving a new message in the chat which hasn't been loade…

seavan:
when receiving a new message in the chat which hasn't been loaded, sorting order is incorrect for notification, specifically in: `chat::onNewMessageLoad`

https://app.clubhouse.io/peerio/story/11879/desktop-old-messages-show-in-desktop-notifications#activity-11954


ebilg:
can be merged

PR #268, opened 2018-08-01
chore: copy for sound selector

suprko:
https://app.clubhouse.io/peerio/story/11724/desktop-notification-sound-settings
 
#### Testing instructions  
n/a, will add desktop PR when it's open

----
### Repository owner

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+src/copy/
--- a/src/copy/en.json
+++ b/src/copy/en.json
@@ -462,6 +462,8 @@
   "title_soundsMention": "You receive a mention or direct messages only",
   "title_soundsMessage": "You receive messages",
   "title_soundsError": "An error occurs",
+  "title_soundSelectHeader": "Notification sound",
+  "title_preview": "Preview",
   "title_desktopNotifications": "Desktop notifications",
   "title_desktopNotificationsDetail": "Show a desktop notification when using the app and...",
   "title_mentionDesktopNotificationsMessage": "You receive a mention or direct message only",


suprko:
note that this PR also includes `.prettierignore` for `src/copy/` folder because Prettier was reformatting `en.json` in an extremely annoying way

anri-asaturov:
Sorting lines or what?

suprko:
linebreak + indent when lines got past a certain length

anri-asaturov:
Could you apply longer printWidth for json files instead, please?
https://prettier.io/docs/en/configuration.html

suprko:
OK I've made an override. I set it to `-1` which disables it. IMO it looks weird when some get indented and some don't, which is inevitable because there are many unbroken strings which go for like 150 characters.

However if we still want it to indent, just at `120` or something, that's easy enough. Personally, I'd rather linebreak+indent all, or none...

lostfictions:
I'd personally recommend not setting a config override, and just turning Prettier off for JSON files in your editor -- it's already configured to not run for JSON files in the project, so adding an override kind of doesn't make sense.

Auto-formatting random JSON files is almost never something you want done automatically in your editor anyway, and if we just leave that on we'll keep running into problems and keep having to add new files and directories to override.

anri-asaturov:
@lostfictions I don't have strong feelings towards this, but I'm curious what is so different about json files in our project compared to js files so they don't need formatting on save? How do you deal with them editing large and multi-level json files like package.json? Indenting manually?
And what kind of problems are we going to keep finding if we auto-format json?

lostfictions:
@anri-asaturov JSON files much more often tend to be at least be partly or entirely machine-generated (think package.json). My experience is that I often find myself opening these files to inspect them or to make a minor change, and then hitting save and having the entire file get churned because of Prettier formatting it. As Lucas was experiencing, it also tries to reformat JSON in a way that ultimately feels inconsistent -- I think the maximum line width rule makes sense for readable code, but arguably just makes highly structured files like JSON less easy to scan by humans. IMO it causes a lot of noise, causes _extra_ noise if or when the file gets rewritten by machine, and is annoying to have to revert via source control. 

More generally my comment in this case is that if we want Prettier to parse JSON files then we might as well turn the lint rule on for them -- I think it's confusing to have an exception for a rule that doesn't exist in the first place. Otherwise, you might as well fix the problem at the editor level, or it's just going to keep happening elsewhere.

anri-asaturov:
Yeah, it's definitely inconsistent now, 
anyway, figure out with Lucas whatever config you think is better in the end, or make a commit here, this branch will get merged soon.

lostfictions:
@suprko up to you on this one -- this is kind of a bikeshed on something relatively minor, so I'd say just do whatever you feel would be easiest/most consistent across our whole codebase.

suprko:
At the very least it's out of scope for this branch to be making changes to Prettier so I've reverted all the changes.

anri-asaturov:
It doesn't save anything back to server

anri-asaturov:
And it's the wrong place to save settings, use Settings keg.

seavan:
@anri-asaturov 
saveProfile will be called in UI to actually send the data to the server.

we discussed with @zippunov that ProfileKeg.payload.props is the best place to save freeform things like this.

saving to SettingsKeg requires modifying the server and schema. is it worth it?

anri-asaturov:
Well it seems illogical to me.
We have Settings keg, which we used up until now to persist settings on server. And this is the only place to store settings on server.
3 month ago we introduced profile props to store medcryptor data, which is exactly user profile properties(medcryptor role etc) and not some random application settings. Now all of the sudden you decide to store next setting in prop?

Right thing to do would be one of the following
1. Remove schema validation from settings keg and store data there
2. Move all settings to profile props, and delete settings keg, at least it's going to be consistent
3. Don't persist sound setting on server, unless we need it for push notifs.

seavan:
ok, then props of for medcryptor/other whitelabel only.

> Remove schema validation from settings keg and store data there
it's not just the schema validation, the data is stored in the flat table. we need to add a field.

> Move all settings to profile props, and delete settings keg, at least it's going to be consistent
wellllll yesss but it requires to rewrite more code on server than I want.

> Don't persist sound setting on server, unless we need it for push notifs.
we need it for push notifs.

In conclusion - I'll add the field to the settings.

seavan:
server PR is here: https://github.com/PeerioTechnologies/peerio-server/pull/724

@anri-asaturov waiting for review.

seavan:
@ebilg ready for testing together with https://github.com/PeerioTechnologies/peerio-server/pull/724 on staging.

PR #276, opened 2018-08-12
feat: uploadTimeFormatted getter

suprko:
https://app.clubhouse.io/peerio/story/2146/client-dates-in-files-are-not-localized
 desktop: https://github.com/PeerioTechnologies/peerio-desktop/pull/406

#### Testing instructions  
see desktop

----
### Repository owner


src/helpers/time-formatting.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Formatted time stamp that changes based on how much time has passed since the given time.
+ * < 24hr passed    => H:MM timestamp
+ * < 48hr passed    => "Yesterday"
+ * < 7 days passed  => day of week
+ * > 7 days passed  => full date
+ */
+function relativeTimestamp(time) {
+    const timeFromNow = Date.now() - time;
```
anri-asaturov: Use minuteClock.now instead of Date.now() here to make it update

suprko:
updated

src/models/files/file.js
```diff
@@ -188,6 +190,20 @@ class File extends Keg {
...
+     * @type {number}
+     */
+    @computed
+    get uploadTimeFormatted() {
+        // If file was uploaded in the past hour, refresh timestamp every minute using `minuteClock` subscription
+        if (this.minuteClock.now - this.uploadedAt < 60 * 60 * 1000) {
+            return relativeTimestamp(this.uploadedAt);
+        }
+        return relativeTimestamp(this.uploadedAt);
+    }
```
seavan: default observableClockEventFrequency is 30, so it's not very much different from `new Clock(60)`

we can probably just use default clock and get rid of the line 201.

what do you think, @anri-asaturov ?

src/models/files/file.js
```diff
@@ -188,6 +190,20 @@ class File extends Keg {
...
+    /**
+     * @type {number}
+     */
+    @computed
+    get uploadTimeFormatted() {
+        // If file was uploaded in the past hour, refresh timestamp every minute using `minuteClock` subscription
+        if (this.minuteClock.now - this.uploadedAt < 60 * 60 * 1000) {
+            return relativeTimestamp(this.uploadedAt);
+        }
+        return relativeTimestamp(this.uploadedAt);
```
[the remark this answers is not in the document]
anri-asaturov: It will always refresh with minuteclock, because there's no code path that does not involve using the observable.
I've commented where you should use the minuteClock.now and that's the best place for it.

src/helpers/time-formatting.js
```diff
@@ -0,0 +1,33 @@
...
+    } else if (timeFromNow > relativeTimeDisplayLimit) {
+        return t('title_yesterday');
+    }
+    return moment(time).fromNow();
+}
+
+module.exports = {
+    relativeTimeDisplayLimit,
+    relativeTimestamp
+};
```
anri-asaturov: Convert to TS please

src/helpers/time-formatting.ts
```diff
@@ -0,0 +1,33 @@
...
+        // Past an hour, refreshing the timestamp every minute is not necessary.
+        timeFromNow = defaultClock.now - time;
+    }
+
+    if (timeFromNow > 7 * day) {
+        return moment(time).format('LL');
+    } else if (timeFromNow > 2 * day) {
+        return moment(time).format('dddd');
+    } else if (timeFromNow > relativeTimeDisplayLimit) {
+        return t('title_yesterday');
```
anri-asaturov: This is based on UTC0 unix timestamp btw, momentjs does take your timezone into account, but `timeFromNow > relativeTimeDisplayLimit` is, strictly speaking, accurate just for people in UTC0 zone. Not sure if it's acceptable, talk to Paul.

src/helpers/time-formatting.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Formatted time stamp that changes baseds on how much time has passed since the given time.
+ * < 24hr passed    => H:MM timestamp
+ * < 48hr passed    => "Yesterday"
+ * < 7 days passed  => day of week
+ * > 7 days passed  => full date
+ */
+export default function relativeTimestamp(time: number): string {
+    let timeFromNow = Date.now() - time;
```
anri-asaturov: Date.now is in utc

ebilg:
commented on ch

src/models/files/file.js
```diff
@@ -1,3 +1,5 @@
+import { relativeTimestamp } from '../../helpers/time-formatting';
```
dchest: `import relativeTimestamp from '../../helpers/time-formatting';` since it's default import.

src/helpers/time-formatting.ts
suprko: Maybe I'm not following the trail right...

`relativeTimeStamp` gets `uploadedAt.valueOf` which is in UTC
`Date.now` is also in UTC.

so `timeFromNow` is accurate. so then `timeFromNow` is used to determine if the file was uploaded an hour ago or a day or a week etc., and *then* UTC time is converted to localized time with `moment`. so... I think we're OK, no?

anri-asaturov: you're right, all good

PR #277, opened 2018-08-14
fix: do not rely on lastMentionID to show mentions [ch12327]

seavan:
https://app.clubhouse.io/peerio/story/11879/desktop-old-messages-show-in-desktop-notifications

I don't really understand the logic behind always scanning for mention in new batch of messages and always displaying only it.

Not only it interferes with the logic of displaying the last message, it causes problems on app start (if the app is yet not loaded).

The other problem I see here is that I couldn't find a way to differentiate between new messages and messages that were loaded as a page when initializing chat.

You always get 84 message kegs or something when receiving only 1 new message.


src/models/chats/chat.js
```diff
@@ -588,7 +588,7 @@ class Chat {
             this.store.onNewMessages({
                 freshBatchMentionCount,
                 lastMessageText,
-                unreadCount: this.unreadCount,
+                unreadCount: 1, // this.unreadCount,
```
seavan: @anri-asaturov this way we get text previews even if we have more than 1 unread message -> much more functional.

anri-asaturov:
@seavan is this considered done or not? I can't say from the code if it's working, might as well approve and send to testing.

seavan:
@anri-asaturov yes it is done

ebilg:
can be merged

PR #278, opened 2018-08-14
feat: username suggestions

alexandra-marin:
 https://app.clubhouse.io/peerio/story/12218/sdk-create-a-function-for-username-suggestions-validated-by-the-server
#### Testing instructions  

----
### Repository owner


src/helpers/validation/user-validators.js
```diff
@@ -152,6 +153,34 @@ const mcrDoctorAhpraAvailability = pair(
...
+
+    for(const option of options) {
+        if(suggestions.length >= maxSuggestions) break;
+        const available = await isValidSignupUsername(option);
+        if(available) {
+            suggestions.push(option);
+        }
+    };
+
+    return suggestions;
```
anri-asaturov: so if i type my first and last name in russian or arabic, what's gonna happen? 
our usernames support english chars only

flohdot: unavailable usernames should not be shown as suggestions. if none of the usernames work, then no suggestions. if this starts being a problem we'll deal with it in a better way.

anri-asaturov: so where's the code that checks all of these for availability?

anri-asaturov: should be within rate limit for now, ok

anri-asaturov: There are some transliteration libs we can use for future.

anri-asaturov:
fix tests please (i recommend installing vscode prettier plugin and/or opting in for commit validation)

PR #279, opened 2018-08-17
feat: pdf generation

anri-asaturov:
platform independent pdf generation

https://github.com/PeerioTechnologies/peerio-desktop/pull/407


src/helpers/pdfform.js.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'pdfform.js' {
```
lostfictions: I'd suggest putting defs at the project root (or in a dedicated subdirectory like `defs/`) instead of scattering them alongside regular code

ebilg:
can be merged

PR #285, opened 2018-08-26
fix: save settings in a concurrent-safe way [ch12811]

seavan:
 
https://app.clubhouse.io/peerio/story/12811/sdk-make-saving-settings-concurrent-safe

because settings keg contain multiple fields which are usually assigned independently, to prevent overwriting values with the new ones from server an updater callback is introduced.

@anri-asaturov it's up to your discretion if we want to get rid of default behaviour (just save all the settings object without updater function) completely, or no.

this is a desktop branch for testing:
https://github.com/PeerioTechnologies/peerio-desktop/tree/feat-settings-concurrent-updates

contains only 1 setting to test - promo emails.

### DIFF
--- a/src/models/user/user.profile.js
+++ b/src/models/user/user.profile.js
@@ -38,7 +38,11 @@ module.exports = function mixUserProfileModule() {
             when(
                 () => !settings.loading,
                 () => {
-                    if (updateFunction) updateFunction(settings);
+                    if (!updateFunction)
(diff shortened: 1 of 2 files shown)


seavan:
@anri-asaturov ready for review w/ clients.

anri-asaturov:
Run prettier to fix test1

ebilg:
can be merged

PR #293, opened 2018-08-29
fix: add files in a single action

dchest:

This relies on the fact that getFileKegsFromServer() will not return duplicate kegs (to avoid searching `filesToAdd` and `filesToRemove` arrays. I'm not sure we can rely on this? 

#### Testing instructions  

Check that files load and deleted files are being deleted.

----
### Repository owner

### DIFF
--- a/src/models/files/file-store-base.js
+++ b/src/models/files/file-store-base.js
@@ -363,8 +363,12 @@ class FileStoreBase {
         }
 
         runInAction(() => {
-            this.knownUpdateId = newKnownUpdateId;
-            this.maxUpdateId = newMaxUpdateId;
+            if (this.knownUpdateId < newKnownUpdateId) {
+                this.knownUpdateId = newKnownUpdateId;
+            }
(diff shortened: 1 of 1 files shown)


src/models/files/file-store-base.js
```diff
@@ -309,60 +309,78 @@ class FileStoreBase {
...
+            // otherwise we insert it into the store
+            if (!existing) {
+                dirty = true;
+                filesToAdd.push([file, keg]);
+            }
+        }
 
+        runInAction(() => {
+            this.knownUpdateId = newKnownUpdateId;
+            this.maxUpdateId = newMaxUpdateId;
```
anri-asaturov: check if for whatever async reason  this.knownUpdateId  is not > than  newKnownUpdateId already.

ebilg:
can be merged

src/models/files/file-store-base.js
dchest: done

PR #297, opened 2018-08-30
fix: hack for progress downloads in case onprogress does not work [ch…

seavan:
…12982]

For mobile on the staging server onprogress event does not fire for some reason yet not clear. On prod it works fine and on desktop it works fine.

My assumption is that something changed with CSP on the file server.

https://app.clubhouse.io/peerio/story/12982/mobile-can-t-download-large-file
 
#### Testing instructions  

mobile and desktop should display adequate download progress when downloading in-chat and in-files.

----
### Repository owner

### DIFF
--- a/src/models/files/file-downloader.js
+++ b/src/models/files/file-downloader.js
@@ -218,6 +218,10 @@ class FileDownloader extends FileProcessor {
                 return true;
             };
 
+            // HACK: if there was no progress event fired
+            // calculate the file progress from completion handler
+            let progressEventFired = false;
+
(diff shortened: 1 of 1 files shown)


src/models/files/file-downloader.js
```diff
@@ -264,6 +271,7 @@ class FileDownloader extends FileProcessor {
             };
 
             xhr.onprogress = function(event) {
+                progressEventFired = true;
```
anri-asaturov: no need for this flag, just set the value on line 246

PR #301, opened 2018-09-03
fix: retry downloads by fetching URL token

dchest:
We only retried the actual XHR request, but we also want to retry chunk
download from the beginning by getting the new chunk URL, since the
failed one could be burned.

Also, add additional check for user-cancelled download.

#### Testing instructions  

Download files (incl. large ones). Verify that they are good.

----
### Repository owner

### DIFF
--- a/src/models/files/file-downloader.js
+++ b/src/models/files/file-downloader.js
@@ -3,6 +3,7 @@ const secret = require('../../crypto/secret');
 const config = require('../../config');
 const FileProcessor = require('./file-processor');
 const { DisconnectedError } = require('../../errors');
+const { retryUntilSuccess } = require('../../helpers/retry');
 
 const { CHUNK_OVERHEAD } = config;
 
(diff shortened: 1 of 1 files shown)


src/models/files/file-downloader.js
```diff
@@ -206,7 +211,7 @@ class FileDownloader extends FileProcessor {
 
             const trySend = () => {
                 // had to do this bcs uploaded blob takes some time to propagate through cloud
-                if (retryCount++ >= 5) return false;
+                if (self.stopped || retryCount++ >= 5) return false;
```
anri-asaturov: but we don't need retryCount anymore, since it's handled by retryUntilSuccess

src/models/files/file-downloader.js
```diff
@@ -234,6 +239,13 @@ class FileDownloader extends FileProcessor {
                     reject(new Error(`Blob download cancelled: ${url}`));
                     return;
                 }
+                if (this.status === 423) {
```
anri-asaturov: ah i see, you want retry on 2 levels

dchest: yup

ebilg:
can be merged

PR #305, opened 2018-09-12
fix: cache existing files

dchest:
Turns out files that are being uploaded do get updated before
descriptors (see https://app.clubhouse.io/peerio/story/13184/), however
they were not cached because they were considered existing.

This commit introduces another array to run in action - filesToCache -
in which all updated files are added, even existing, so that they can be
cached.

Fixes https://app.clubhouse.io/peerio/story/13184/cachedescriptor-not-found-after-uploading-a-file

#### Testing instructions  

A.

1. Upload a file into Files.
2. Make sure there's no "cacheDescriptor was called, but cached keg not found" error in console.

B.

1. Clear cache and log in — make sure there are no strange or cache-related errors in console.
2. Check in devtools if cache is filled.

----
### Repository owner


ebilg:
can be merged

PR #306, opened 2018-09-12
Fixes fixes fixes

lostfictions:

Some bugfixes:

- Bulk folder delete was dumping its contents back into the root instead of deleting them (e51ec60)
- `pickPathSelector` was not being awaited (a88cbd7)
- Model types weren't actually being exported from `peerio-icebear/models` (`export * from './file-name' ` can't re-export defaults since they don't have names!) (8561fab)

Type fixes:
- `File` and `FileFolder` can now be discriminated via `isFolder` field, but only via strict comparison: `item.isFolder === true` (this is a quirk since we don't have `strictNullChecks` on -- `null` and `undefined` being in every type are hindering inference via just eg. `if(item.isFolder)`)
- A few extra annotations, some fields and methods marked as `protected`

#### Testing instructions  

----
### Repository owner


ebilg:
can be merged

PR #316, opened 2018-09-20
chore: bulk send telemetry events

suprko:
https://app.clubhouse.io/peerio/story/13804/sdk-telemetry-bulksend-function
desktop: https://github.com/PeerioTechnologies/peerio-desktop/pull/433

#### Testing instructions  
see desktop

----
### Repository owner


src/telemetry/main.ts
```diff
@@ -45,38 +48,43 @@ interface BaseProperties {
...
+        if (!User.current) {
+            telemetryStore.save(eventObj);
+            return;
+        }
 
+        const baseProperties = await getBaseProperties();
+        const eventProperties = convertEventPropertyCase(eventObj);
         eventObj.properties = { ...baseProperties, ...eventProperties };
 
         const data = g.btoa(JSON.stringify(eventObj));
```
lostfictions: @anri-asaturov is there already a `btoa` equivalent defined somewhere in icebear or am i misremembering?

anri-asaturov: yeah, crypto/util/conversion
`bytesToB64(strToBytes(...))`

src/telemetry/store.ts
```diff
@@ -0,0 +1,12 @@
...
+
+class TelemetryStore {
+    events: EventObject[] = [];
+
+    save = (ev: EventObject) => {
+        this.events.push(ev);
+    };
+}
+
+const store = new TelemetryStore();
```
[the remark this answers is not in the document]
anri-asaturov: TBH would be enough to just declare an array in main.ts where this data is actually used.
I assume 'sendStored' is getting called from some observable reaction in desktop, and that's also unnecessary, that reaction could live in main.ts too.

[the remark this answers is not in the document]
lostfictions: I was thinking if Lucas had the impulse to generalize this, it's because he thought there might be more instances where it might be useful? I'm not too sure where -- I can't really think of other places where telemetry would need to be "atomic" once a user has given their consent (since in most scenarios we'd want to send incremental events, even for cancellation, etc.) but it doesn't seem out of the question.

But yeah- if it can be even simpler, then... maybe it should be :)

suprko:
`TelemetryStore` is toast. just an array in `main.ts` now, very simple.

Currently, desktop calls `telemetry.sendStored` on login or signup completion. We can make this call in icebear if that's preferred? Just in the login `_postAuth` block?

lostfictions:
@suprko is this update reflected in https://github.com/PeerioTechnologies/peerio-desktop/pull/433/files ?

lostfictions:
Ah wait, misunderstood, I thought you meant the store was just an array in Desktop now

lostfictions:
Just noting I commented on this at https://github.com/PeerioTechnologies/peerio-desktop/pull/433#issuecomment-426792859 -- starting to have trouble reviewing/keeping track across multiple repos

src/telemetry/main.ts
```diff
@@ -86,3 +95,38 @@ export async function send(eventObj: EventObject): Promise<void> {
...
+        });
+        // console.log(events);
+
+        // Mixpanel accepts batch events but up to a maximum of 50 "messages". This term is not explained.
+        // We are guessing here. Currently assuming that "message" corresponds to an entire "event".
+        // Increment can be adjusted as we learn more.
+        const increment = 50;
+        for (let i = 0; i < events.length; i += increment) {
+            const chunk = events.slice(i, i + increment);
+            const data = g.btoa(JSON.stringify(chunk));
```
lostfictions: Ultimately not sure if it matters whether we use `btoa` or `bytesToB64` but we should probably stick to one for consistency. @anri-asaturov is there a preference?

src/telemetry/main.ts
```diff
@@ -86,3 +95,38 @@ export async function send(eventObj: EventObject): Promise<void> {
...
+
+            g.fetch(config.telemetry.baseUrl, {
```
lostfictions: actually, I'm not totally sure what the idea behind this `g` is... can someone clarify? shouldn't this just use an existing isofetch module that polyfills/proxies so that this can run on node as well as browser?

src/telemetry/main.ts
```diff
@@ -86,3 +95,38 @@ export async function send(eventObj: EventObject): Promise<void> {
...
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/x-www-form-urlencoded'
+                },
+                body: `data=${data}`
+            });
+
+            console.log(chunk);
```
lostfictions: I think I flagged this log already in a previous PR...? (Or a different repo? Sorry...) Wasn't the conclusion to remove it before this gets merged?

src/telemetry/main.ts
```diff
@@ -86,3 +95,38 @@ export async function send(eventObj: EventObject): Promise<void> {
...
+export async function sendStored(): Promise<void> {
+    try {
+        const baseProperties = await getBaseProperties();
+        const events = eventStore;
+
+        events.forEach(ev => {
+            const eventProperties = convertEventPropertyCase(ev);
+            ev.properties = { ...baseProperties, ...eventProperties };
+        });
+        // console.log(events);
```
lostfictions: cleanup commented debugs plz

src/telemetry/main.ts
```diff
@@ -45,38 +47,45 @@ interface BaseProperties {
...
+        const eventProperties = convertEventPropertyCase(eventObj);
         eventObj.properties = { ...baseProperties, ...eventProperties };
 
-        const data = g.btoa(JSON.stringify(eventObj));
-        const url = `${config.telemetry.baseUrl}${data}`;
+        const data = bytesToB64(strToBytes(JSON.stringify(eventObj)));
+        const url = `${config.telemetry.baseUrl}?data=${data}`;
 
         console.log(eventObj);
         g.fetch(url, {
```
lostfictions: see below question about just using isofetch :)

src/telemetry/main.ts
```diff
@@ -45,38 +47,45 @@ interface BaseProperties {
...
+
+    Object.keys(event.properties).forEach(itemInCamel => {
+        const item = camelToTitleCase(itemInCamel);
+        eventProperties[item] = event.properties[itemInCamel];
+    });
+
+    return eventProperties;
 }
 
+let eventStore: EventObject[] = [];
```
lostfictions: can this technically grow without bound if the user is idling at the login screen? might seem minor, but it'd kind of suck if this was a memory leak for the person who runs peerio at launch but has auto-login disabled.

probably a minor concern but we might still want to observe the length and start dequeueing events (or just stop accepting them) if it grows too large. seems like something that might easily happen by accident, especially via some unrelated future change.

suprko: what would be a reasonable upper bound for this kinda thing?

seavan: @suprko let's put it in SDK config and default to 500.

as it reaches the limit, each event pushed to the end of the array should remove the first element (FIFO-style).

lostfictions:
@suprko okay, i've pushed some fixes -- i took a cursory look that it's reacting correctly, but i don't have your workflow for verifying login or checking on the mixpanel side, so please verify that it all looks good.

`fetch` isn't available in node (and i'm not sure about the story in react-native) so i've added `cross-fetch` so that this PR will keep working across environments. i also added some `await`s where they seemed to be missing.

suprko:
confirmed that it's working on the mixpanel side

src/telemetry/main.ts
```diff
@@ -45,44 +48,93 @@ interface BaseProperties {
...
-        const eventProperties: EventProperties = {};
-        Object.keys(eventObj.properties).forEach(itemInCamel => {
-            const item = camelToTitleCase(itemInCamel);
-            eventProperties[item] = eventObj.properties[itemInCamel];
-        });
+    if (!User.current) {
+        eventStore.push(eventObj);
+        return;
+    }
+    if (User.current && !User.current.settings.dataCollection) return;
```
anri-asaturov: note that `User.current.settings.dataCollection` is `false` up until user is authenticated and `User.current.settings.loaded === true`. So all events generated after User.current has been assigned and up until settings are loaded will get discarded. Is that the expected behaviour?

src/telemetry/main.ts
anri-asaturov: I usually try to estimate the average RAM taken by one event and think about how much RAM can I afford to reserve for this.

src/telemetry/main.ts
```diff
@@ -45,44 +48,93 @@ interface BaseProperties {
...
+}
+
+autorun(() => {
+    if (
+        User.current &&
+        User.current.settings.loaded &&
+        User.current.settings.dataCollection &&
+        eventStore.length > 0
+    ) {
+        sendStored();
```
anri-asaturov: not sure if `autorun` is needed here, we don't want sendStore() triggered every time `dataCollection` changes and `eventStore` never changes after events are sent. `when` will be more appropriate here, I try to avoid autorun that triggers important (data mutating/sending) events, as it is easy to miss when it triggers unnecessarily.

src/telemetry/main.ts
anri-asaturov: Also remember that User.current instance(and username) can change until finally authenticated

suprko: This is supposed to mean "if current user exists but user has not consented to dataCollection, do not send/store events"... Is that not the case? `User.current` shouldn't be assigned until signup and login has succeeded, no?

src/telemetry/main.ts
```diff
@@ -45,44 +48,93 @@ interface BaseProperties {
...
+
+        eventStore.forEach(ev => {
+            const eventProperties = convertEventPropertyCase(ev);
+            ev.properties = { ...baseProperties, ...eventProperties };
+        });
+
+        // Mixpanel accepts batch events but up to a maximum of 50 "messages". This term is not explained.
+        // We are guessing here. Currently assuming that "message" corresponds to an entire "event".
+        // Increment can be adjusted as we learn more.
+        const increment = 50;
```
anri-asaturov: extract to a `MIXPANEL_BATCH_SIZE` constant in the top of this file pls, with a comment not to exceed 50

src/telemetry/main.ts
anri-asaturov: You have just reworked login/signup recently :-P
User.current gets assigned before User.current.login() is called.
Every login attempt assigns User.current a new instance.
https://github.com/PeerioTechnologies/peerio-desktop/blob/dev/src/ui/login/Login.jsx#L154

anri-asaturov: Also note that events from different(unauthenticated) users might get collected, every failed attempt changes the current user.

anri-asaturov:
> Ah wait, misunderstood, I thought you meant the store was just an array in Desktop now. Did you disagree that callers should just keep track of their own arrays and `sendStored` should just accept an array?
> 
> To me this code is still more stateful than I'd like and doesn't fix one of the fundamental problems I described... If you think the alternative would be really clunky I'm open to alternative suggestions though.

Just saw you comment from 7 days ago, @lostfictions , no wonders with the all-hands busyness we had.

So we have a problem of events not being able to be sent to mixpanel at the moment they've been generated but they're safe to be sent as soon as it gets possible. Why do you think clients should be solving this problem given that solution in every client will be the same?

lostfictions:
@anri-asaturov talked this through with lucas. it's resolved

src/telemetry/main.ts
anri-asaturov: Although this will get merged with offline queue anyway I assume, and there we will want to react differently, nevermind.

lostfictions: It's effectively the same in this case, except that `when` will fail in the scenario where a user signs out and then creates a new account.

anri-asaturov: signout always reloads the app

lostfictions: if you'd like, i'll put `eventStore.length > 0` to the top of the predicate so that it short-circuits quicker?

anri-asaturov: Not worth the trouble, you'll be merging this with offline queue right? Or it's gonna be separate?

lostfictions: oops, was replying to your first comment, github didn't reload all the intermediate ones

anri-asaturov:
CI is failing but it's a server (i hope) bug with invites, which i'm debugging atm

PR #340, opened 2018-10-23
chore: do not exclude deleted contacts from keg

suprko:
 https://app.clubhouse.io/peerio/story/15413/desktop-deleted-accounts-broken

#### Testing instructions  

----
### Repository owner

### DIFF
--- a/src/models/kegs/shared-db-boot-keg.ts
+++ b/src/models/kegs/shared-db-boot-keg.ts
@@ -336,7 +336,7 @@ class SharedDbBootKeg extends SyncedKeg<ISharedDbBootKegPayload, ISharedDbBootKe
                 keys: {}
             };
             this.participants.forEach(c => {
-                if (c.isDeleted || c.notFound) return;
+                if (c.notFound) return;
                 if (c.loading) {
                     throw new Error(
                         `Can not save boot keg because participant Contact (${


src/models/kegs/shared-db-boot-keg.ts
```diff
@@ -336,7 +336,7 @@ class SharedDbBootKeg extends SyncedKeg<ISharedDbBootKegPayload, ISharedDbBootKe
                 keys: {}
             };
             this.participants.forEach(c => {
-                if (c.isDeleted || c.notFound) return;
+                if (c.notFound) return;
```
suprko: @anri-asaturov 
The problem is that deleted users get excluded from `boot`. So when you open a DM with a user who is deleted, a chat is created with only you as the participant; the deleted user does not show up as a participant. So then the app treats this DM as a `self` DM.

This change allows `isDeleted` contacts back into the boot, which allows deleted contacts to be added to a new DM, so this fixes the issue described in Clubhouse. But... is this gonna cause huge problems elsewhere? Since `isDeleted` contacts are being filtered out here I'm not really sure where they might need to be filtered out downstream instead.

anri-asaturov: 👍 +1

PR #354, opened 2018-11-09
chore: small type fixes from quick audit

lostfictions:
#### Testing instructions  

----
### Repository owner

### DIFF
--- a/src/crypto/keys.ts
+++ b/src/crypto/keys.ts
@@ -1,7 +1,7 @@
 /**
  * Peerio Crypto module for key handling.
  */
-import BLAKE2s, { ByteArray } from 'blake2s-js';
+import BLAKE2s from 'blake2s-js';
 import * as nacl from 'tweetnacl';
 import { scryptPromise } from './scrypt-proxy';
 import { bytesToHex, strToBytes, concatTypedArrays, getRandomBytes } from './util';
@@ -143,10 +143,8 @@ export function generateAuthSalt(): Uint8Array {
  * Hashes auth public key. Uses personalized hash.
  * @returns 32 bytes personalized hash
  */
-export function getAuthKeyHash(key: ByteArray): Uint8Array {
-    const hash = new BLAKE2s(32, {
-        personalization: strToBytes('AuthCPK1')
-    });
+export function getAuthKeyHash(key: Uint8Array): Uint8Array {
+    const hash = new BLAKE2s(32, { personalization: strToBytes('AuthCPK1') });
     hash.update(key);
     return hash.digest();
 }
--- a/src/crypto/keys.ts
+++ b/src/crypto/keys.ts
@@ -1,7 +1,7 @@
 /**
  * Peerio Crypto module for key handling.
  */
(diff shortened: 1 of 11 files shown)


src/crypto/keys.ts
```diff
@@ -145,7 +143,7 @@ export function generateAuthSalt(): Uint8Array {
  * Hashes auth public key. Uses personalized hash.
  * @returns 32 bytes personalized hash
  */
-export function getAuthKeyHash(key): Uint8Array {
+export function getAuthKeyHash(key: ByteArray): Uint8Array {
```
anri-asaturov: `getAuthKeyHash` is actually called with Uint8Array argument, not sure why blake2s-js defines its own ByteArray

lostfictions: it's an alias, i think, for u8arr | number[], so just u8arr should be safe. i'll double-check

src/crypto/secret.ts
```diff
@@ -126,7 +126,7 @@ export function decrypt(
     containsLength?: boolean
 ): Uint8Array {
     let start = 0;
-    let end: number;
+    let end: number | undefined;
```
anri-asaturov: shouldn't this be `let end?: number`? :-P

anri-asaturov:
```suggestion
    let end?: number;
```
let's test how suggestions work :)

src/crypto/util/random.ts
```diff
@@ -14,7 +14,7 @@ import { InvalidArgumentError } from '../../errors';
  * @returns random bytes array of `num` size
  */
 // eslint-disable-next-line import/no-mutable-exports
-export let getRandomBytes: (num: number) => Uint8Array;
+export let getRandomBytes!: (num: number) => Uint8Array;
```
anri-asaturov: Oh so `!` here will affect any imported `getRandomBytes` and you won't need to use `!` in the place where you import it? nice

lostfictions: yeah `!` is two kinds of asserts: (1) asserting something you're about to use is non-null; (2) asserting that a binding (either a `let` or a class field) is "definitely assigned", as in definite assignment analysis. obviously the latter is ideal where possible, since if you can ensure something is definitely assigned when declaring it, it can save a lot of "i promise this isn't null now" asserts at callsites.

src/crypto/secret.ts
lostfictions: i... think it worked, let's see if ci likes it

lostfictions: haha, yeah -- this isn't actually valid syntax here. good to know that accepted suggestions can't actually be reverted, though....

lostfictions: (it's for sure confusing that that's allowed for class fields and function parameters but not let-bindings)

dchest: It kinda makes sense: in class fields and function arguments `name?` means it's optional, that is there might not even be `name` in this context, even though it's no different from it being `undefined`, but `let name` declares that `name` is definitely here, i.e. there's no syntax in JS to make it optional.

PR #359, opened 2018-11-21
feat: website unfurl

dchest:
 
https://app.clubhouse.io/peerio/story/17304/sdk-parse-opengraph-and-twitter-card-meta-tags

desktop: https://github.com/PeerioTechnologies/peerio-desktop/pull/500

mobile: https://github.com/PeerioTechnologies/peerio-mobile/pull/380

#### Testing instructions  

Try messaging with various website and image links. Follow desktop testing instructions (https://github.com/PeerioTechnologies/peerio-desktop/pull/500)

----
### Repository owner


src/helpers/unfurl/parse.ts
```diff
@@ -76,7 +76,7 @@ export function parseHTML(url: string, data: string): HTMLParseResult | null {
         });
 
         // Validate and trim strings.
-        res.siteName = trimOrUndefined(res.title);
+        res.siteName = trimOrUndefined(res.siteName);
```
suprko: `siteName` was getting overwritten by `title` and I was sure this was the problem but it's still happening when I test... :/

dchest: Thanks, indeed. You probably have it cached, do  `ice.CacheEngineBase.clearAllCache()` in console.

package.json
```diff
@@ -51,6 +52,7 @@
     "mkdirp": "0.5.1",
     "mobx": "4.5.2",
     "moment": "2.22.2",
+    "parse5": "5.1.0",
```
anri-asaturov: Does this work with mobile?
Why not try regex? :trollface:

src/config.ts
```diff
@@ -298,15 +298,24 @@ export class Config {
...
 
-    /**
-     * How long to wait for external server to respond when unfurling urls posted in messages.
-     */
-    unfurlTimeout = 30000;
+    unfurl = {
+        /**
+         * Maximum number of links in message to unfurl.
+         */
+        maxLinks: 64,
```
anri-asaturov: I think this is too much, were there any requirements on this?

package.json
dchest: Yes, it should work.

src/config.ts
dchest: Nope. Previously we didn't limit this at all, AFAIK. What would you suggest?

anri-asaturov: Let's ask product team, i suspect they will have a number under 3.

anri-asaturov: @paulrdukes

dchest: FYI, this number includes both image and website links.

paulrdukes: I'm currently fine with all of them until someone complains. 😬 If someone is going to send 50 links in a single message, complain to them though. Not to us. 😐

dchest: Each website link sends 2-3 requests, each image link 1 request. So, if the message has too many links, it can make Peerio slooow. So we should have some limit. Although, of course, someone can just send many messages, but then our rate limits on server should kick in.

paulrdukes: What constitutes too many links? Is that based on user connection? User hardware? Would things speed up as links are processed? What does slow Peerio look like? Is it a lag in response to user interactions?

anri-asaturov: We'd like to know that all in such detail too, Paul, but it's impossible without wasting the time on creating test cases and measuring. Imagine you open and load 10 browser tabs with 10 different urls at the same time, that's what happens when your Peerio encounters a message with 10 links in it. Now imagine this is a mobile browser.

anri-asaturov: Additionally, the urls will not be expanded in the rendered message until processing is done.

paulrdukes: > Additionally, the urls will not be expanded in the rendered message until processing is done.

@seavan is this read receipts on mobile all over again?

seavan: @paulrdukes we will cache url info locally, and it doesn't update after being loaded, so no

dchest:
YouTube sucks, doesn't give any info in og or description/title. But at least we have API that we can use later when we have custom parsers.
screen shot 2018-11-27 at 15 20 57

dchest:
unWIP'd and ready for review

suprko:
Certain URLs aren't parsing and I can't figure out why. e.g. this one:
https://support.peerio.com/hc/en-us/articles/360018254751-How-to-organise-your-files-in-Peerio

doesn't show any preview content at all. `XMLHttpRequest` seems to be failing?

suprko:
Favicons often missing, see Mona's comment here https://app.clubhouse.io/peerio/story/17377/desktop-url-preview#activity-17521

src/config.ts
```diff
@@ -298,15 +299,36 @@ export class Config {
...
+         * How long to wait for external server to respond when unfurling urls posted in messages.
+         */
+        timeout: 30000,
+
+        /**
+         * Maximum HTML content length to fetch for parsing.
+         * If larger, will be truncated.
+         * Note that it's the number of UTF-16 characters, not bytes (same for ASCII).
+         */
+        maxHTMLContentLength: 64 * 1024, // 64K ought to be enough for anyone,
```
lostfictions: @dchest i see what you did there :wink:

src/helpers/unfurl/parse.ts
```diff
@@ -0,0 +1,140 @@
...
+    title?: string;
+    description?: string;
+    imageURL?: string;
+    imageAlt?: string;
+    faviconURL?: string;
+}
+
+export function parseHTML(url: string, data: string): HTMLParseResult | null {
+    try {
+        const doc = parse5.parse(data) as any; // parse5 typing isn't good
```
lostfictions: Was wondering if it'd be better to use the native `DOMParser` where available, but presumably the perf benefit is too small for the added complexity of implementation differences?

lostfictions:
lgtm at a first pass, not having tried it yet

src/helpers/unfurl/parse.ts
dchest: Then we'll have to have two parsers, since this one doesn't return DOM. parse5 is pretty slow, but we never parse more than 65K anyway.

dchest:
@lostfictions yep, good idea about problematic cases. I'll think of something.

ebilg:
can be merged

PR #362, opened 2018-11-27
fix: don't try to fetch from IP or local domains; don't unfurl non-200 status

dchest:
Last-minute fixes!

#### Testing instructions  

----
### Repository owner


dchest:
Just in case someone has something running on `localhost` and then someone sends them: `https://localhost:9999/req=DROP ALL TABLES`

PR #367, opened 2018-12-03
fix: optimize chat list loading [ch17654]

dchest:

https://app.clubhouse.io/peerio/story/17654/fix-one-by-one-loading-and-redrawing-of-chat-list
 
#### Testing instructions  

1. Make sure chats load.
2. On Android, check if initial chat list loading is not loading rooms one-by-one as it was before.

----
### Repository owner


dchest:
@seavan I have a suspicion that mobile doesn't wait for `chatStore.loaded` (or `!chatStore.loading`) during the loading screen. I've checked but couldn't find the place. Could you please check it?

src/models/chats/chat-store.ts
```diff
@@ -145,6 +146,7 @@ export class ChatStore {
      */
     @computed
     get channels() {
+        if (!this.loaded) return [];
```
dchest: Before this during initial load it was recalculating `channels` every time the new chat head loads, triggering re-render. Now it waits until `this.loaded` before it returns all loaded chats. Note that we have a timeout to waiting before `this.loaded` is true in `loadAllChats`, so it won't wait for all chats to load if it takes more than  5 seconds.

anri-asaturov: yeah this is fine

seavan: @dchest yes it does wait during the loading screen:

it waits for chatState.init which waits for chatStore.loaded here:

https://github.com/PeerioTechnologies/peerio-mobile/blob/chore-login-refactor/app/components/messaging/chat-state.js#L53

